PR #27, opened 2018-07-31
obey the multiaddr's IP version

### LINKED ISSUE #24: Obey the multiaddr's IP version when listening and dialing
Multiaddrs explicitly include the IP version so we need to make sure to listen/dial with `udp#` instead of just `udp`.

### LINKED ISSUE #25: Crate a separate "dialer" connection for both IPv6 and IPv4
We can't dial an IPv4 addr from an IPv6 UDP connection and vice versa.

marten-seemann:
Fixes #24. Fixes #25. Depends on #26.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,6 +18,10 @@ before_install:
   - export GOARCH=$TRAVIS_GOARCH
   - go env # for debugging
 
+# see https://github.com/travis-ci/travis-ci/issues/8361#issuecomment-350090030
+before_script:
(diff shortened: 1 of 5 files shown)


Stebalien:
FYI, take a look at https://github.com/libp2p/go-tcp-transport/blob/master/.travis.yml. We have a "standard" CI script that takes care of fixing travis, vetting, etc.

Stebalien:
Got it. We currently just run the other ginkgo test from go tests.

PR #28, opened 2018-08-01
fix aggregation of coverage profiles

marten-seemann:


marten-seemann:
While technically not necessary, since we don't have subpackages (and also no subpackages 2 levels deep), we might as well as merge it since it's better than we have so far.

marten-seemann:
And Codecov didn't kick in when I force-pushed this PR. I don't want to waste time on debugging this now. Closing.

PR #29, opened 2018-08-05
generate certificates that are valid for 180 days

### LINKED ISSUE #22: think about certificate expiry
Certificates are currently valid for 30 days.

marten-seemann:
Fixes #22.

By setting the `NotBefore` and `NotAfter` in the certificate, we're leaking the exact start time of a node. I'm not sure if we're doing that at other places as well, but if not, we could add some randomisation here.

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -18,6 +18,8 @@ import (
 // mint certificate selection is broken.
 const hostname = "quic.ipfs"
 
+const certValidityPeriod = 180 * 24 * time.Hour
+
 func generateConfig(privKey ic.PrivKey) (*tls.Config, error) {
 	key, hostCert, err := keyToCertificate(privKey)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


whyrusleeping:
@marten-seemann if a node happens to be online long enough for this to expire, do they get any sort of log messages indicating whats going wrong?

marten-seemann:
Unfortunately, TLS alerts are really unhelpful (in most cases, you just get something like "bad certificate"). Furthermore, since it's the peer that's making the decision to reject a certificate, the node owner will not get any helpful error message.

Any suggestions how to handle that? Should we just increase the duration even more?

Stebalien:
Could we have the config's `GetCertificate` and `GetClientCertificate` regenerate certs on-demand when necessary.

In the future, if/when we add support for "offline" key material, we'd have to setup some warning/async system but we can handle that later.

marten-seemann:
Yeah, I thought about that, but mint (the TLS 1.3 implementation quic-go currently uses) doesn't support `GetCertificate` and `GetClientCertificate`.

I'd like to merge this PR soon, and find a better solution later. Users who activate experimental features usually upgrade their nodes frequently, so I don't expect that we'll run into problems if we roll this out with a long enough validity period.

Stebalien:
Fine by me.

whyrusleeping:
yep, SGTM, ship it!

Issue #8, opened 2018-06-12
Dial from a port we're listening on


Stebalien:
QUIC should try to pick a port we're listening on when dialing instead of opening a new one. We may want to extract some of the logic in https://github.com/libp2p/go-reuseport-transport.

marten-seemann:
I’m not exactly sure how the reuseport-transport works, but I think there’s an even better solution with QUIC. Since QUIC connections are identified by a connection ID, we can use a single port for all outgoing connections.

I’ve been working on a set of changes to make this possible in quic-go (https://github.com/lucas-clemente/quic-go/issues/1324), and I’m expecting to merge these next week or so.

Stebalien:
Nice!

So, my point about reuseport-transport is that it can handle the case where we're listening on *multiple* ports. That is, users may listen on `localhost:123` and `my_public_addr:345`. We'd want to use port 345 when dialing out. reuseport-transport has a bunch of logic for picking a reasonable source port based on the destination address.

marten-seemann:
Ok, that makes sense.
We should use that logic once quic-go supports running a listener and a dialer on the same packet conn (https://github.com/lucas-clemente/quic-go/issues/561). It's a kind of complicated issue, so I don't expect to fix it within the next few weeks.

Stebalien:
Thinking through this a bit, the approach we're currently using in go-reuseport-transport won't work. If a libp2p node is configured to listen on some unicast IP X and then tries to dial a unicast IP Y, that may fail with both #34 and #44 as there may be no route from X to Y. go-reuseport-transport works around this by:

1. Picking the best *port* it can by looking at address types.
2. Binding to `0.0.0.0:PORT` instead of `X:PORT`.
3. Dialing.

The trick is dialing from 0.0.0.0 instead of X. Unfortunately, we can't reuse this same trick for QUIC because we don't want to open a new file descriptor bound to 0.0.0.0.

---

The only solution I can think of is to do this the *right* way and use [netlink](https://godoc.org/github.com/vishvananda/netlink#RouteGet). See: https://github.com/libp2p/go-reuseport-transport/issues/2

A partial solution would be to reuse the source port *if* and only if we're listening on 0.0.0.0.

Stebalien:
It may also be possible to use `/proc/net/route` along with the list of interfaces. Unfortunately, `netlink` will only work on Linux.

[...]

PR #34, opened 2018-10-17
Reuse UDP port when possible

cannium:
Try to mimic the behavior of libp2p/go-reuseport-transport

Fix https://github.com/libp2p/go-libp2p-quic-transport/issues/8


listener.go
```diff
@@ -99,6 +92,15 @@ func (l *listener) setupConn(sess quic.Session) (tpt.Conn, error) {
 
 // Close closes the listener.
 func (l *listener) Close() error {
+	network := l.conn.LocalAddr().Network()
+	l.transport.connManager.mu.Lock()
+	switch network {
+	case "udp4":
```
Stebalien: So, go has a habit of throwing away ip6/ip4 information and returning just "udp". We probably need to *remember* whether we're using IPv4/IPv6 in the listener itself.

transport.go
```diff
@@ -32,27 +32,73 @@ var quicConfig = &quic.Config{
...
+func newConnManager() *connManager {
+	return &connManager{
+		ipv4Conns: make(map[net.PacketConn]int),
+		ipv6Conns: make(map[net.PacketConn]int),
+	}
 }
 
-func (c *connManager) GetConnForAddr(network string) (net.PacketConn, error) {
+func pickDialConn(conns map[net.PacketConn]int, remoteHostIP net.IP) net.PacketConn {
+	backup := make([]net.PacketConn, 0, len(conns))
```
Stebalien: So, this is going to be called *a lot* (every dial). Let's just set *one* backup if and only if a backup hasn't been set. We don't really need to collect a list of them and then select the first.

transport.go
```diff
@@ -32,27 +32,73 @@ var quicConfig = &quic.Config{
...
-		return c.connIPv4, err
+		conn := pickDialConn(c.ipv4Conns, remoteAddr.IP)
+		if conn == nil {
+			conn, err = c.createConn(network, "0.0.0.0:0")
+			if err != nil {
+				return nil, err
+			}
+		}
+		c.mu.Lock()
+		c.ipv4Conns[conn] += 1
```
Stebalien: What are we using these refcounts for?

cannium: My initial thought was not actually close a socket when upper layer call `Close()`, except the refcount down to 0. But `go-reuseport-transport` doesn't do this, so I'd like to discuss if it's reasonable.

transport.go
cannium: Nice, will do.

listener.go
cannium: Since key of map is the connection itself, call `delete` for both `ipv4Conns` and `ipv6Conns` might be easier.

cannium:
From my test, in current usage pattern, `listen` comes before `dial`, but yes, that could be a problem. So it's another trade-off to deal, if it's possible that `dial` comes before `listen`, a change is necessary.

cannium:
ping @marten-seemann for review

transport.go
```diff
@@ -32,27 +32,72 @@ var quicConfig = &quic.Config{
...
+	}
+
 	switch network {
 	case "udp4":
-		var err error
-		c.connIPv4Once.Do(func() {
-			c.connIPv4, err = c.createConn(network, "0.0.0.0:0")
-		})
-		return c.connIPv4, err
+		conn := c.pickDialConn(c.ipv4Conns, remoteAddr.IP)
```
Stebalien: This is going to need some synchronization (we're reading a shared ipv4Conns map and writing a shared `c.defaultConn` variable.

Stebalien:
> So it's another trade-off to deal, if it's possible that dial comes before listen, a change is necessary.

It's unusual but completely possible.

cannium:
> > So it's another trade-off to deal, if it's possible that dial comes before listen, a change is necessary.
> 
> It's unusual but completely possible.

So I track connections like `conn -> conn addr type` in `connManager`,

cannium:
After some trial and error, it occurs to me that the right map for connections is `conn -> remote addr type`, which could also be interpreted as `conn -> usage of the conn`, so "conn A is used for local peers", "conn B is used for global peers", and "conn C is the default conn, for unspecified peers".

listener.go
```diff
@@ -99,6 +92,22 @@ func (l *listener) setupConn(sess quic.Session) (tpt.Conn, error) {
 
 // Close closes the listener.
 func (l *listener) Close() error {
+	connManager := l.transport.connManager
+	network := l.conn.LocalAddr().Network()
+	connManager.mu.Lock()
+	switch network {
+	case "udp4":
+		delete(connManager.ipv4Conns, l.conn)
```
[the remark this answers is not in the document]
cannium: I believe `l.quicListener.Close()` would call `PacketConn`'s `Close()`, so the connection could not be used anymore.

transport.go
```diff
@@ -31,28 +31,88 @@ var quicConfig = &quic.Config{
...
+	unspecified addrType = 3
+)
+
 type connManager struct {
-	connIPv4Once sync.Once
-	connIPv4     net.PacketConn
+	mu          sync.Mutex
+	defaultConn net.PacketConn
+	// map underhood PacketConn -> connection remote address type(usage of the conn)
+	ipv4Conns map[net.PacketConn]addrType
```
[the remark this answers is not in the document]
cannium: Then the map would become `map[addrType][]net.PacketConn`, which is actually a little more complex than current map.

transport.go
```diff
@@ -66,6 +126,25 @@ func (c *connManager) createConn(network, host string) (net.PacketConn, error) {
 	return net.ListenUDP(network, addr)
 }
 
+func (c *connManager) listenUDP(addr ma.Multiaddr) (net.PacketConn, error) {
+	network, host, err := manet.DialArgs(addr)
+	if err != nil {
+		return nil, err
+	}
+	conn, err := c.createConn(network, host)
```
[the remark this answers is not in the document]
cannium: As mentioned by @Stebalien [here](https://github.com/libp2p/go-reuseport-transport/pull/3#issuecomment-427183029):

> We can reuse the same port (and even the same underlying OS socket) by calling pc := net.ListenPacket("udp", ":NNNN"). From here, we can send packets to any ip/port with pc.WriteTo(buf, dest) and read incoming packets from any ip/port with pc.ReadFrom(buf).

OS kernel handles it for us. 
Also, very rarely we need to listen on ports dynamically.

Stebalien:
Let's go back to the drawing board with this. Let's keep the code but close the PR so we can spend some time figuring out the right way to proceed. As-is, the current approach isn't going to work. See: https://github.com/libp2p/go-libp2p-quic-transport/issues/8#issuecomment-452105642

PR #35, opened 2018-11-08
Export Conn.Context()

lthibault:
I find myself needing to do cleanup logic when connections are closed.  When using https://github.com/lucas-clemente/quic-go, this is easily doable using the exported `Context()` methods on `quic.Stream` and `quic.Session`.

This change does **not** change the public api, but allows users to recover the `Context()` methods using type assertions, e.g.:

```go
type Contexter interface{
    Context() context.Context
}

// ... 
cx := someLibp2pConn.(Contexter).Context()

// ... 
```

See also https://github.com/libp2p/go-libp2p-net/issues/32 & https://github.com/libp2p/go-libp2p-net/issues/33

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1,6 +1,8 @@
 package libp2pquic
 
 import (
+	"context"
+
 	ic "github.com/libp2p/go-libp2p-crypto"
 	peer "github.com/libp2p/go-libp2p-peer"
 	tpt "github.com/libp2p/go-libp2p-transport"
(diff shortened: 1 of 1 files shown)


marten-seemann:
@Stebalien, what do you think about this? Does it make sense to expose this before https://github.com/libp2p/go-libp2p-net/issues/32?

Stebalien:
Revisiting this, this isn't going to work. Libp2p will _wrap_ the underlying connection, so there's no way to make this work unless libp2p re-exports this function.

PR #37, opened 2018-11-15
only use a single certificate [do not merge]

### LINKED ISSUE #21: Custom length certificate changes
Currently, we require length two chains where the bottom key (the one used directly in the handshake) is ephemeral. However, RSA signature verification is a bit expensive, especially when rapidly forming many connections (e.g., in the DHT).

It would be nice if we could instead make this configurable:

* By default, we'd start out with this two key system.
* By default, we'd allow anywhere from 1 to 4 certificates in the chain (restricted to 4 to prevent clients from tricking servers into validating long certificate chains).

Motivation:

* Certificate verification can be expensive (depending on the algorithm).
* In the future, peers will likely have an outer "ephemeral" peer ID (generated on boot) and will likely negotiate additional sub-peer IDs later (per application). When using these ephemeral peer IDs, the current two-certificate system doesn't provide any additional security.

### LINKED ISSUE #23: Explicitly note certificate "purpose" (the subject)
Currently, if a peer uses their peer ID to sign a certificate, QUIC will allow the signed key to act on behalf of the peer. The peer should have to *explicitly state* that they want to delegate their peer ID.

Motivation: If we don't do this, we won't be able to use peer IDs to sign any certificates *except* delegation certificates.

Proposal: Put `/p2p/QmPeerID` in the certificate's subject field (or some other field if that doesn't make sense). The certificate would only be valid if there exists some certificate higher up the chain signed by that peer ID.

marten-seemann:
Closes #21. Closes #23.

We don't have to worry about backwards compatibility **if** we make sure to not merge this PR before using a new QUIC version. 

**Please DON'T merge!**

### DIFF
--- a/crypto.go
+++ b/crypto.go
@@ -1,8 +1,6 @@
 package libp2pquic
 
 import (
-	"crypto/ecdsa"
-	"crypto/elliptic"
 	"crypto/rand"
 	"crypto/tls"
 	"crypto/x509"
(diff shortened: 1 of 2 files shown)


crypto.go
```diff
@@ -21,29 +19,7 @@ const hostname = "quic.ipfs"
 const certValidityPeriod = 180 * 24 * time.Hour
```
Stebalien: We might as well just make this infinite, at this point.

marten-seemann:
Closing due to age. #67 contains the updated handshake logic.

PR #43, opened 2018-12-12
Migrate to go-libp2p-tls

### LINKED ISSUE #18: reuse code from go-libp2p-tls
Given that QUIC is using the standard golang TLS config type, it would be awesome if we could extract this code out into a new library and then re-use the same handshake for TLS over TCP. Something like:

```go
type Identity tls.Config

func NewIdentity(privKey ic.PrivKey) (*Identity, error) { return generateConfig(privKey) }

// for dialing
func (i *Identity) ConfigForPeer(remote ic.PubKey) (*tls.Config, error) {
  // clone the config and set the validator.
}

func KeyFromChain([]*x509...) (ic.PubKey, error) { ... }
```

(and any other useful helper functions you can think of)

Ichbinjoe:
Directly closes #18 and https://github.com/libp2p/go-libp2p-tls/issues/4


Ichbinjoe:
Blocked by https://github.com/libp2p/go-libp2p-tls/pull/7 as well as a go-conn-security version bump (conflicting types)

Ichbinjoe:
This fails since its dependencies were not updated. On my local system with test dependency this passes. This is still blocked on libp2p/go-libp2p-tls#7.

I removed two tests since they no longer make sense - the server isn't able to validate the client ID against the certificate it receives, so it doesn't make sense to check to be sure who connected to us is someone we _expected_ to connect to us, since that concept doesn't exist anymore.

In addition, this behavior can probably be tested instead in the go-libp2p-tls package in the future.

marten-seemann:
This is obsolete now that https://github.com/libp2p/go-libp2p-tls/pull/20 has landed.

PR #44, opened 2018-12-22
enabling same port for dialling and listening

upperwal:
Idea is to create ```PacketConn``` once either during Dialling or Listening and reusing it for the rest of the program lifetime. ```GetConnForAddr``` is responsible to either return a ```PacketConn``` for a specific network type or create it (only once) if not available.

@Stebalien Tested this with chat and echo example and it's working. Let me know if this looks ok to you.

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -26,20 +26,16 @@ type listener struct {
 
 var _ tpt.Listener = &listener{}
 
-func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID, key ic.PrivKey, tlsConf *tls.Config) (tpt.Listener, error) {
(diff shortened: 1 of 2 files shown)


upperwal:
failing tests

Stebalien:
You should coordinate with @cannium. He's been working on https://github.com/libp2p/go-libp2p-quic-transport/pull/34 (currently waiting for a review from @marten-seemann...).

The tricky part is handling the case where the user listens on localhost and then tries dialing a non-localhost address.

Stebalien:
Let's go back to the drawing board with this. Let's keep the code but close the PR so we can spend some time figuring out the right way to proceed. As-is, the current approach isn't going to work. See: https://github.com/libp2p/go-libp2p-quic-transport/issues/8#issuecomment-452105642

upperwal:
@Stebalien 

> Let's go back to the drawing board with this

Yeah, now I can see how different interfaces could be a problem.

PR #47, opened 2019-02-13
Add go mod support + adjust Travis

anacrolix:
The unit tests pass with go mod locally, but CircleCI seems to freeze up during the tests. Is there something extra the CI might require to finish @marten-seemann ?


.circleci/config.yml
```diff
@@ -0,0 +1,10 @@
+version: 2
+jobs:
+  build:
+    docker:
+      - image: circleci/golang:1.11
+        environment:
+          GO111MODULE: "on"
+    steps:
+      - checkout
+      - run: go test -v -race -bench . ./...
```
marten-seemann: Better to use ginkgo instead of go test:
```
ginkgo -r -v --randomizeAllSpecs --randomizeSuites --trace --progress
```

anacrolix:
I've added them together, because they synergize.

.circleci/config.yml
anacrolix: How do you run benchmarks with gingko?

anacrolix:
So the tests fail for IPv6, I've had this elsewhere with Circle CI.

marten-seemann:
There’s a fix for that in the Travis config. You can probably just copy that line.

anacrolix:
Unfortunately it was a lot more involved than that, but I got it working.

raulk:
@marten-seemann 

> I still don't understand why we need go mod support immediately. Adding go mod support when most upstream libraries aren't supporting yet it seems to create a lot of hassle for us.

We need a way to make QUIC support available in libp2p for downstream consumers that don't use gx (the majority). All of libp2p, IPFS and multiformats are adopting gomod now, so rather than plain ol' vendoring, we should adopt gomod here too.

Good thing is that we can reference a specific commit from the gomod file, and if quic-go doesn't want to add gomod yet (btw - what is the reason to defer until go 1.12?), we can find a way around it.

marten-seemann:
To make this repo go-gettable, we just need to merge #42. That PR has been waiting for review for weeks (months?) now...

anacrolix:
What's the hold up? Let's get this merged.

marten-seemann:
I'd like to do this without CircleCI.
 I'm working on a ginkgo-enabled Travis config that builds using both gx and gomod in https://github.com/libp2p/go-libp2p-tls/pull/19, and I intend to apply a similar config here.

anacrolix:
Happy to merge the go-mod part only then?

marten-seemann:
Why not merge all in one? I prefer to merge changes that are properly tested on CI.

anacrolix:
@marten-seemann https://github.com/libp2p/go-libp2p-quic-transport/pull/47#pullrequestreview-204546640.

marten-seemann:
It's our policy to only merge changes when they have proper tests, so I think it's reasonable to block this until https://github.com/libp2p/go-libp2p-tls/pull/19 is done. We have downstream users that use gx and others that use go get, so the responsible thing to do is testing both install paths.

I don't think I'm contradicting my earlier statement in https://github.com/libp2p/go-libp2p-quic-transport/pull/47#pullrequestreview-204546640 here. I was just pointing out that adding go mod support and adding a CircleCI config are orthogonal to each other, and should have been dealt with in two separate PR.

raulk:
I pushed a commit to fix the gomod file. I'm going to rm the CircleCI config as it's incomplete anyway (only tests gomod, not gx), and I'm updating the Travis config with the one submitted by @marten-seemann here: https://github.com/libp2p/go-libp2p-tls/pull/19.

At some point we should try to converge towards the unified Travis config that uses Travis build matrices. Example: https://github.com/libp2p/go-libp2p-kad-dht/blob/master/.travis.yml. Not sure if we can make it work with ginkgo.

raulk:
All set here.

Tried asking you for a review @anacrolix, but Github doesn't show you in the dropdown ¯\_(ツ)_/¯

Kubuxu:
Wait with merge as I am currently rewriting tags of `ipfs` org which will break tag of `ipfs/go-log`.

raulk:
@Kubuxu ack.

anacrolix:
Note that `go-log` is now broken for go mod, because tag changes.

anacrolix:
I have stopped CircleCI from building this repo.

marten-seemann:
As @Kubuxu said...
> Wait with merge as I am currently rewriting tags of `ipfs` org which will break tag of `ipfs/go-log`.

@anacrolix Can you please fix this, I don't like to have repositories in a state where you can't rely on CI.

anacrolix:
Oops I missed that comment.

PR #52, opened 2019-03-20
Quic reuseport

### LINKED ISSUE #8: Dial from a port we're listening on

lnykww:
Fixes #8 
Implementation：https://github.com/libp2p/go-libp2p-quic-transport/issues/8#issuecomment-472653841

I am not sure where the source select library should be placed, so I created a repository on my github to put this code. If it is not appropriate, please let me know where to put this code.

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -16,9 +16,8 @@ var quicListenAddr = quic.ListenAddr
 
 // A listener listens for QUIC connections.
 type listener struct {
-	quicListener quic.Listener
-	transport    tpt.Transport
-
+	quicListener   quic.Listener
+	transport      *transport
 	privKey        ic.PrivKey
 	localPeer      peer.ID
 	localMultiaddr ma.Multiaddr
@@ -26,7 +25,7 @@ type listener struct {
 
 var _ tpt.Listener = &listener{}
 
-func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID, key ic.PrivKey, tlsConf *tls.Config) (tpt.Listener, error) {
+func newListener(addr ma.Multiaddr, t *transport, localPeer peer.ID, key ic.PrivKey, tlsConf *tls.Config) (tpt.Listener, error) {
 	lnet, host, err := manet.DialArgs(addr)
 	if err != nil {
 		return nil, err
@@ -35,7 +34,7 @@ func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID, 
 	if err != nil {
 		return nil, err
 	}
-	conn, err := net.ListenUDP(lnet, laddr)
+	conn, err := t.connManagers.Listen(lnet, laddr)
 	if err != nil {
 		return nil, err
 	}
@@ -49,7 +48,7 @@ func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID, 
 	}
 	return &listener{
 		quicListener:   ln,
-		transport:      transport,
+		transport:      t,
 		privKey:        key,
 		localPeer:      localPeer,
 		localMultiaddr: localMultiaddr,
@@ -99,6 +98,18 @@ func (l *listener) setupConn(sess quic.Session) (tpt.Conn, error) {
 
 // Close closes the listener.
 func (l *listener) Close() error {
+	lnet, host, err := manet.DialArgs(l.localMultiaddr)
+	if err != nil {
+		return err
+	}
+	laddr, err := net.ResolveUDPAddr(lnet, host)
+	if err != nil {
+		return err
+	}
+	err = l.transport.connManagers.Close(lnet, laddr)
+	if err != nil {
+		return err
+	}
 	return l.quicListener.Close()
 }
 
--- a/listener.go
+++ b/listener.go
@@ -34,7 +34,7 @@ func newListener(addr ma.Multiaddr, t *transport, localPeer peer.ID, key ic.Priv
 	if err != nil {
 		return nil, err
 	}
-	conn, err := t.ConnManagers.Listen(lnet, laddr)
+	conn, err := t.ConnManagers().Listen(lnet, laddr)
 	if err != nil {
 		return nil, err
 	}
@@ -106,7 +106,7 @@ func (l *listener) Close() error {
 	if err != nil {
 		return err
 	}
-	err = l.transport.ConnManagers.Close(lnet, laddr)
+	err = l.transport.ConnManagers().Close(lnet, laddr)
 	if err != nil {
 		return err
 	}
--- a/listener.go
+++ b/listener.go
@@ -34,7 +34,7 @@ func newListener(addr ma.Multiaddr, t *transport, localPeer peer.ID, key ic.Priv
 	if err != nil {
 		return nil, err
 	}
-	conn, err := t.connManagers.Listen(lnet, laddr)
+	conn, err := t.ConnManagers.Listen(lnet, laddr)
 	if err != nil {
 		return nil, err
 	}
@@ -106,7 +106,7 @@ func (l *listener) Close() error {
 	if err != nil {
 		return err
 	}
-	err = l.transport.connManagers.Close(lnet, laddr)
+	err = l.transport.ConnManagers.Close(lnet, laddr)
 	if err != nil {
 		return err
 	}
--- a/reuse.go
+++ b/reuse.go
@@ -1,0 +1,190 @@
+package libp2pquic
+
+import (
+	"errors"
+	"net"
+	"sync"
+
+	srcs "github.com/lnykww/go-src-select"
+)
+
+type ReuseConn struct {
+	net.PacketConn
+	lock sync.Mutex
+	ref  int
+}
+
+func NewReuseConn(conn net.PacketConn) *ReuseConn {
+	return &ReuseConn{
+		PacketConn: conn,
+		ref:        1,
+		lock:       sync.Mutex{},
+	}
+}
+
+func (rc *ReuseConn) Ref() error {
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+	if rc.ref == 0 {
+		return errors.New("conn closed")
+	}
+	rc.ref++
+	return nil
+}
+
+func (rc *ReuseConn) Close() error {
+	rc.lock.Lock()
+	defer rc.lock.Unlock()
+	var err error
+	switch rc.ref {
+	case 0: // cloesd, just return
+		return nil
+	case 1: // no reference, close the conn
+		err = rc.PacketConn.Close()
+	}
+	rc.ref--
+	return err
+}
+
+type Reuse struct {
+	lock           sync.Mutex
+	unicast        map[string]map[int]net.PacketConn
+	unspecific     []net.PacketConn
+	connGlobal     net.PacketConn
+	connGlobalOnce sync.Once
+}
+
+func NewReuse() *Reuse {
+	return &Reuse{
+		unicast:    make(map[string]map[int]net.PacketConn),
+		unspecific: make([]net.PacketConn, 0),
+	}
+}
+
+// getConnGlobal get the global random port socket, if not exist, create
+// it first.
+func (r *Reuse) getConnGlobal(network string) (net.PacketConn, error) {
+	var err error
+	r.connGlobalOnce.Do(func() {
+		var addr *net.UDPAddr
+		var conn net.PacketConn
+		var host string
+		switch network {
+		case "udp4":
+			host = "0.0.0.0:0"
+		case "udp6":
+			host = "[::]:0"
+		}
+		addr, err = net.ResolveUDPAddr(network, host)
+		if err != nil {
+			return
+		}
+		conn, err = net.ListenUDP(network, addr)
+		if err != nil {
+			return
+		}
+
+		r.connGlobal = NewReuseConn(conn)
+	})
+	if r.connGlobal == nil && err == nil {
+		err = errors.New("global socket init not done")
+	}
+	return r.connGlobal, err
+}
+
+// rueseConn Assertion the type of the conn is ReuseConn and inc the ref
+func (r *Reuse) reuseConn(conn net.PacketConn) error {
+	reuseConn, ok := conn.(*ReuseConn)
+	if !ok {
+		panic("type ReuseConn Assert failed: something wrong!")
+	}
+	return reuseConn.Ref()
+}
+
+func (r *Reuse) dial(network string, raddr *net.UDPAddr) (net.PacketConn, error) {
+	// Find the source address which kernel use
+	sip, err := srcs.Select(raddr.IP)
+	if err != nil {
+		return r.getConnGlobal(network)
+	}
+	r.lock.Lock()
+	defer r.lock.Unlock()
+
+	// If we has a listener on this address, use it to dial
+	if c, ok := r.unicast[sip.String()]; ok {
+		for _, v := range c {
+			return v, nil
+		}
+	}
+
+	if len(r.unspecific) != 0 {
+		return r.unspecific[0], nil
+	}
+
+	return r.getConnGlobal(network)
+}
+
+func (r *Reuse) Dial(network string, raddr *net.UDPAddr) (net.PacketConn, error) {
+	conn, err := r.dial(network, raddr)
+	if err != nil {
+		return nil, err
+	}
+	// we are reuse a conn, reference it
+	if err = r.reuseConn(conn); err != nil {
+		return nil, err
+	}
+	return conn, nil
+}
+
+func (r *Reuse) Listen(network string, laddr *net.UDPAddr) (net.PacketConn, error) {
+	conn, err := net.ListenUDP(network, laddr)
+	if err != nil {
+		return nil, err
+	}
+
+	reuseConn := NewReuseConn(conn)
+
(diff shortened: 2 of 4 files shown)


transport.go
```diff
@@ -66,12 +129,32 @@ func (c *connManager) createConn(network, host string) (net.PacketConn, error) {
 	return net.ListenUDP(network, addr)
 }
 
+func (c *connManagers) AddListener(network string, addr *net.UDPAddr, conn net.PacketConn) error {
+	if connManager, ok := c.managers[network]; ok {
+		return connManager.AddListener(network, addr, conn)
```
Stebalien: Nit: Feeding the network through here just to be thrown away seems funny. We don't _have_ to use the exact same interface for the connManager and the connManagers.

(although I don't mind much either way)

listener.go
```diff
@@ -49,10 +54,12 @@ func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID,
 	}
 	return &listener{
 		quicListener:   ln,
-		transport:      transport,
+		t:              t,
```
Stebalien: Let's just keep this as `transport`.

listener.go
```diff
@@ -39,6 +40,10 @@ func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID,
 	if err != nil {
 		return nil, err
 	}
+	err = t.connManagers.AddListener(lnet, laddr, conn)
+	if err != nil {
```
Stebalien: Do we need to close the connection?

listener.go
```diff
@@ -16,17 +16,18 @@ var quicListenAddr = quic.ListenAddr
...
-	quicListener quic.Listener
-	transport    tpt.Transport
-
+	quicListener   quic.Listener
+	t              *transport
 	privKey        ic.PrivKey
 	localPeer      peer.ID
 	localMultiaddr ma.Multiaddr
+	udpAddr        *net.UDPAddr
+	network        string
```
Stebalien: I'd mildly prefer to just re-parse the multiaddr on close (storing duplicate information bugs me).

lnykww:
> Nice!
 >
 > * This will need a bunch of tests. Especially around closing listeners, connections, picking the right one, etc. They don't have to be _that_ fancy, but we should test unspecified versus localhost, etc.
 
Test Added.

 > * We can leave your dep where it is or adopt it into the libp2p org, I'm happy to do either. I'd also like to try adding windows support before shipping this but that shouldn't actually be _too_ difficult (we can work on that after we finish up with    this PR).
 
Adopt it into the libp2p org will be better.

 > The current design (as well as the current code, for that matter) is nice and simple but, as far as I can tell, it doesn't allow us to actually _close_ the underlying udp sockets when we no longer need them (because we can't _know_ if we no longer      need them). We may want to reference count on create/close.
 >
 > If we're going to be wrapping connections anyways, we should consider a manager of the form:
 >
 > ```go
 > type Manager interface {
 >     // Dial dials the remote addr, reusing an existing socket if possible.
 >     Dial(network string, addr *net.UDPAddr) (net.PacketConn, error)
 >     // Listen listens on the given address.
 >     Listen(network string, addr *net.UDPAddr) (net.PacketConn, error)
 > }
 > ```
 >
 > Where Dial/Listen return connections wrapped in a reference counter.
 >
 > * Closing any connection (listen/dial) would decrement the reference count. When it hits zero, the underlying connection would be closed.
 > * Closing a listener would also remove the listener from the list of open listeners.
 >
 > That way, we no longer need to manually register/unregister connections. The connection manager just does everything for us.
 >
 > Note: I don't feel strongly about this interface and, IMO, the current interface is just fine. My main concern is reference counting.
 >
 > If you'd like, I'd be fine punting reference counting given that it's an existing issue.

Used referenced connection now.

@Stebalien  I have updated the code with your review opinoin. Please for a second review.

reuse.go
```diff
@@ -0,0 +1,190 @@
+package libp2pquic
+
+import (
+	"errors"
+	"net"
+	"sync"
+
+	srcs "github.com/lnykww/go-src-select"
```
Stebalien: I've forked this to github.com/libp2p/go-src-select I can also delete that and we can transfer it if you'd like.

Could you add a LICENSE? Ideally Apache 2.0 + MIT. Take a look at the COPYRIGHT, LICENSE-MIT, and LICENSE-APACHE files in https://github.com/ipfs/go-ds-crdt/commit/d51c9f1306f317b77e7c314113b8643a0a471b82.

Note: This project is still copyright _you_ (well, you and whoever else ends up contributing to it).

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+)
+
+type ReuseConn struct {
```
Stebalien: I'd keep all these types internal for now.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+// rueseConn Assertion the type of the conn is ReuseConn and inc the ref
+func (r *Reuse) reuseConn(conn net.PacketConn) error {
+	reuseConn, ok := conn.(*ReuseConn)
+	if !ok {
+		panic("type ReuseConn Assert failed: something wrong!")
+	}
+	return reuseConn.Ref()
+}
+
+func (r *Reuse) dial(network string, raddr *net.UDPAddr) (net.PacketConn, error) {
```
Stebalien: We can return a `*ReuseConn` here. Then we can drop `reuseConn` and just call `Ref` directly.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+
+func NewReuseConn(conn net.PacketConn) *ReuseConn {
+	return &ReuseConn{
+		PacketConn: conn,
+		ref:        1,
+		lock:       sync.Mutex{},
+	}
+}
+
+func (rc *ReuseConn) Ref() error {
```
Stebalien: Probably simpler just to return a boolean `true` for success.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+	return r.getConnGlobal(network)
+}
+
+func (r *Reuse) Dial(network string, raddr *net.UDPAddr) (net.PacketConn, error) {
+	conn, err := r.dial(network, raddr)
+	if err != nil {
+		return nil, err
+	}
+	// we are reuse a conn, reference it
+	if err = r.reuseConn(conn); err != nil {
```
Stebalien: * I believe this will double-ref if we end up creating a new global connection.
* We probably need a loop (either inside dial out outside). Otherwise, we could try to use the connection but fail because some other thread decided it didn't want it. In that case, we should just open a new connection.

Stebalien:
So, I realized something annoying. quic-go correctly won't close connections it didn't create. @marten-seemann is this something that could be made configurable? I.e., some way to "give" a connection to QUIC? This patch is introducing connection refcounting but nothing will ever call `Close`.

@lnykww otherwise, we'll have to stash the `pconn` in the `go-libp2p-quic-transport.conn` and close it manually.

Let's wait for @marten-seemann to chime in here before we move forward.

reuse.go
lnykww: Done

Stebalien: Thanks!

reuse.go
lnykww: > * I believe this will double-ref if we end up creating a new global connection.

I have found this problem before. But we create a listen socket as a global socket, 
it means that if we don't want to use it, we need close it ourself in this module, 
rather than the user in other module. So when it was created, the reference should be 1. 

> * We probably need a loop (either inside dial out outside). Otherwise, we could try to use the connection but fail because some other thread decided it didn't want it. In that case, we should just open a new connection.

Ok, i will add a retry for that.

Stebalien: > I have found this problem before. But we create a listen socket as a global socket,
it means that if we don't want to use it, we need close it ourself in this module,
rather than the user in other module. 

My thinking was that, when the last connection using the global connection closes, we could close the global connection (reopening a new one when necessary).

lnykww: > My thinking was that, when the last connection using the global connection closes, we could close the global connection (reopening a new one when necessary).

This may cause us to create new socket frequently.

Stebalien: If we're repeatedly closing all connections. Also note: this will only happen if we aren't listening on a relevant ip address.

My worry here is that we're _leaking_ a socket. In theory, a user should be able to spin up/down new libp2p instances without leaking anything.

lnykww: Yes, you are right. It does not happen very often. I will do that.

listener.go
```diff
@@ -35,7 +34,7 @@ func newListener(addr ma.Multiaddr, transport tpt.Transport, localPeer peer.ID,
 	if err != nil {
 		return nil, err
 	}
-	conn, err := net.ListenUDP(lnet, laddr)
+	conn, err := t.connManagers.Listen(lnet, laddr)
```
marten-seemann: Please don't access unexported fields of other structs.

listener.go
```diff
@@ -99,6 +98,18 @@ func (l *listener) setupConn(sess quic.Session) (tpt.Conn, error) {
...
 func (l *listener) Close() error {
+	lnet, host, err := manet.DialArgs(l.localMultiaddr)
+	if err != nil {
+		return err
+	}
+	laddr, err := net.ResolveUDPAddr(lnet, host)
+	if err != nil {
+		return err
+	}
+	err = l.transport.connManagers.Close(lnet, laddr)
```
marten-seemann: Same here.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+	"errors"
+	"net"
+	"sync"
+
+	srcs "github.com/lnykww/go-src-select"
+)
+
+type ReuseConn struct {
+	net.PacketConn
+	lock sync.Mutex
```
marten-seemann: Convention: This should be called `mutex`.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+	ref  int
+}
+
+func NewReuseConn(conn net.PacketConn) *ReuseConn {
+	return &ReuseConn{
+		PacketConn: conn,
+		ref:        1,
+		lock:       sync.Mutex{},
```
marten-seemann: No need to initialize a mutex. The default value works fine.

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+	"net"
+	"sync"
+
+	srcs "github.com/lnykww/go-src-select"
+)
+
+type ReuseConn struct {
+	net.PacketConn
+	lock sync.Mutex
+	ref  int
```
marten-seemann: Why is `ref` an `int`, if it only ever takes the value `0` or `1`?

reuse.go
```diff
@@ -0,0 +1,190 @@
...
+		return nil
+	case 1: // no reference, close the conn
+		err = rc.PacketConn.Close()
+	}
+	rc.ref--
+	return err
+}
+
+type Reuse struct {
+	lock           sync.Mutex
```
marten-seemann: Same here.

reuse.go
lnykww: ref indicates how many connection was use it. and its value is not only  0 or 1.

listener.go
marten-seemann: I should have been more specific. Please don't access **any** fields of other structs.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+type Reuse struct {
+	mutex      sync.Mutex
+	unicast    map[string]map[int]*reuseConn
+	unspecific []*reuseConn
+	connGlobal *reuseConn
+}
+
+func NewReuse() *Reuse {
+	return &Reuse{
+		unicast:    make(map[string]map[int]*reuseConn),
```
marten-seemann: I assume the `int` should be the port number, right? This will need documentation here.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+
+	switch {
+	case laddr.IP.IsUnspecified():
+		r.unspecific = append(r.unspecific, rconn)
+	default:
+		if _, ok := r.unicast[laddr.IP.String()]; !ok {
+			r.unicast[laddr.IP.String()] = make(map[int]*reuseConn)
+		}
+		if _, ok := r.unicast[laddr.IP.String()][laddr.Port]; ok {
+			conn.Close()
```
marten-seemann: Why are we closing the connection that we just created a few lines above? Can't we avoid creating it in the first place?

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func NewReuseConn(conn net.PacketConn) *reuseConn {
+	return &reuseConn{
+		PacketConn: conn,
+		ref:        1,
+	}
+}
+
+func (rc *reuseConn) Ref() bool {
```
marten-seemann: Things will become a lot clearer if you introduce a function to tell if `rc.ref > 0` and another function to increase the value.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+}
+
+func (r *Reuse) Listen(network string, laddr *net.UDPAddr) (net.PacketConn, error) {
+	conn, err := net.ListenUDP(network, laddr)
+	if err != nil {
+		return nil, err
+	}
+
+	rconn := NewReuseConn(conn)
+	_ = rconn.Ref()
```
marten-seemann: This should not be necessary any more with the refactoring I suggested for the ref counter.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+
+	switch {
```
marten-seemann: This switch statement would greatly benefit from comments explaining what's going on here.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+	}
+	return rconn, nil
+}
+
+func (r *Reuse) Close(addr *net.UDPAddr) error {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	switch {
+	case addr.IP.IsUnspecified():
+		for index, conn := range r.unspecific {
```
marten-seemann: Can we use a `map` for `r.unspecific`? Looping over a slice and then deleting values from the middle of it doesn't look very clean.

reuse.go
```diff
@@ -0,0 +1,187 @@
...
+			recAddr := conn.LocalAddr().(*net.UDPAddr)
+			if recAddr.IP.Equal(addr.IP) && recAddr.Port == addr.Port {
+				r.unspecific = append(r.unspecific[:index], r.unspecific[index+1:]...)
+				return nil
+			}
+		}
```
marten-seemann: What if the connection is not found in the slice? We should handle that case and return an error.
The code will probably become more readable if you use an `if` instead a `switch`.

reuse.go
lnykww: it is a bug here. it need not to increase the ref. Because of NewReuseConn already set the ref to 1. Will fix it.

reuse.go
lnykww: emm, I think we need not to check if we has already listening on an address. Because when we call ListenUDP, the kernel will do that.

Stebalien:
@marten-seemann can you comment on:

> So, I realized something annoying. quic-go correctly won't close connections it didn't create. @marten-seemann is this something that could be made configurable? I.e., some way to "give" a connection to QUIC? This patch is introducing connection refcounting but nothing will ever call Close.

?

marten-seemann:
This is behaving exactly as intended. If you're passing a connection to quic-go, either in `quic.Listen` or `quic.Dial`, you can reuse the same connection for multiple connections, and quic-go will handle the multipling. quic-go will therefore never close this connection, it's your responsibility to do this.
If you use `quic.ListenAddr` or `quic.DialAddr`, quic-go creates a new connection just for this one purpose, and closes this connection as soon as it's done with it.

Stebalien:
I understand, I'm asking if we could have some way to "give" the connection to quic-go. Alternatively, we can handle the reference counting from the go-libp2p-transport Conn.

marten-seemann:
The point is that quic-go doesn't know when it's ok to close the connection. Only the user knows when it's done `quic.Dial`ing on the connection. Why is this a problem here in the first place?

Stebalien:
The current patch refcounts connections, incrementing the refcount whenever it hands one to quic.

(Note: I'm asking if some kind of "owned" flag would be a reasonable addition to quic-go, I know it won't work at the moment)

marten-seemann:
I'm sorry, I'm afraid I don't understand the problem you're trying to solve. When and why do you want to close a UDP connection?

Stebalien:
When we, e.g., close a listener. We should be able to spin up a bunch of libp2p nodes in as single process and then tear them down without leaking anything.

marten-seemann:
Then we need to close the underlying connections in the same step as we close the listeners. Note that this will also kill all outgoing connections we might have use the packet conn for.

Stebalien:
That's why this is doing ref-counting.

lnykww:
> I'm sorry, I'm afraid I don't understand the problem you're trying to solve. When and why do you want to close a UDP connection?
@marten-seemann  For creating a new libp2p instance, we might create a new global conn. When we close the libp2p instance, we need to close the global conn to prevent the connection from leaking.
> The point is that quic-go doesn't know when it's ok to close the connection. Only the user knows when it's done quic.Dialing on the connection. Why is this a problem here in the first place?
Did this means that we need close the connection ourselves?
@Stebalien

PR #54, opened 2019-04-04
update quic-go to v0.11.0

### LINKED ISSUE #53: Update quic-go
To fix the deadlock. See: https://github.com/ipfs/go-ipfs/issues/6168.

Ideally, we'd get this into the impending go-ipfs release.

marten-seemann:
Fixes #53.
Release-blocker (?) for https://github.com/ipfs/go-ipfs/issues/6163, so we should merge this asap.


stream.go
```diff
@@ -12,8 +12,7 @@ type stream struct {
 var _ smux.Stream = &stream{}
 
 func (s *stream) Reset() error {
-	if err := s.Stream.CancelRead(0); err != nil {
-		return err
-	}
-	return s.Stream.CancelWrite(0)
+	s.Stream.CancelRead(0)
+	s.Stream.CancelWrite(0)
+	return nil
```
Stebalien: Any reason not to return these?

marten-seemann: There’s nothing to return. They don’t have a return value any more.

Stebalien: Got it.

PR #59, opened 2019-04-20
when ListenUDP fails once, try again next time

### LINKED ISSUE #58: GetConnForAddr returns nil conn after ListenUDP failed

marten-seemann:
Fixes #58.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -31,26 +31,30 @@ var quicConfig = &quic.Config{
 }
 
 type connManager struct {
-	connIPv4Once sync.Once
-	connIPv4     net.PacketConn
+	mutex sync.Mutex
 
-	connIPv6Once sync.Once
-	connIPv6     net.PacketConn
+	connIPv4 net.PacketConn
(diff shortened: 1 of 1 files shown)


transport.go
```diff
@@ -31,26 +31,30 @@ var quicConfig = &quic.Config{
...
 		return c.connIPv4, err
 	case "udp6":
+		if c.connIPv6 != nil {
+			return c.connIPv6, nil
+		}
 		var err error
-		c.connIPv6Once.Do(func() {
-			c.connIPv6, err = c.createConn(network, ":0")
-		})
+		c.connIPv6, err = c.createConn(network, ":0")
```
anacrolix: Why not `[::]:0`?

marten-seemann: I'm not sure. `:0` seems to work fine. Do you see any advantage in changing this?

anacrolix: No I was just curious. As long as "udp6" is specified, it shouldn't matter.

PR #63, opened 2019-06-02
implement connection reuse

### LINKED ISSUE #8: Dial from a port we're listening on

marten-seemann:
Fixes #8. Closes #52.

This PR builds on #52, cleans up a lot of code, and makes sure that connections that we're listening on are actually reused for dialing (that was a bug in #52).

The logic is now the following:

* For calls to `Listen`, always create a new connection.
  * on 0.0.0.0: call this a "global" connection
  * on any other IP: call this a "unicast" connection
* When dialing
  * if netlink is available (i.e. on Linux), look up the source IP address that the kernel would choose. If we have a unicast connection, use that one
  * otherwise, take any global connection (we don't care about the port number)
  * if no global connection is available, create one

The listening part of this is still a bit racy, if the user decides to listen on an ephemeral port that was picked for a `Dial` call before. We should reuse connections when `Listen` is called after `Dial` (and the port happens to be the same) as well. Planning to fix this in a separate PR.

As far as I understand, on non-Linux systems, this is effectively equivalent to what we have now: we use the same connection, which is listening on 0.0.0.0, for all outgoing dials. @Stebalien, is that what you described in https://github.com/libp2p/go-libp2p-quic-transport/issues/8#issuecomment-472653841? I'd like to hear your feedback on this PR before continuing the work.

This PR does not YET deal with closing connections. The `reuseConn` is already reference counted, so we can always get the number of QUIC connections that are running on top of this connection. However, we don't do anything once this number reaches 0.
There are two ways we can handle this:
1. Introduce a callback, and immediately close the connection when the reference counter reaches 0.
2. Implement a sweeper that runs periodically, and close connections that have been idle for longer than x.

I'm leaning towards 2., since this will make sequential dials (which might happen for reconnects) more efficient.


marten-seemann:
@Stebalien Any thoughts on this PR?

Stebalien:
> As far as I understand, on non-Linux systems, this is effectively equivalent to what we have now: we use the same connection, which is listening on 0.0.0.0, for all outgoing dials. @Stebalien, is that what you described in #8 (comment)? I'd like to hear your feedback on this PR before continuing the work.

Sounds like it, I'll read the code.

> I'm leaning towards 2., since this will make sequential dials (which might happen for reconnects) more efficient.

I agree.

reuse.go
```diff
@@ -0,0 +1,140 @@
...
+		unicast: make(map[string]map[int]*reuseConn),
+		global:  make(map[int]*reuseConn),
+	}
+}
+
+func (r *reuse) getSourceIPs(network string, raddr *net.UDPAddr) ([]net.IP, error) {
+	// Determine the source address that the kernel would use for this IP address.
+	// Note: This only works on Linux.
+	// On other OSes, this will return a netlink.ErrNotImplemetned.
+	routes, err := (&netlink.Handle{}).RouteGet(raddr.IP)
```
Stebalien: We have to _create_ (constructor) the handle (and close it later). We should probably share one for the "reuse" instance.

reuse.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return ips, nil
+}
+
+func (r *reuse) Dial(network string, raddr *net.UDPAddr) (*reuseConn, error) {
+	conn, err := r.dial(network, raddr)
+	if err != nil {
+		return nil, err
+	}
+	conn.IncreaseCount()
```
Stebalien: Race?

reuse.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	if err != nil {
+		return nil, err
+	}
+	conn, err := net.ListenUDP(network, addr)
+	if err != nil {
+		return nil, err
+	}
+	rconn := newReuseConn(conn)
+	r.global[conn.LocalAddr().(*net.UDPAddr).Port] = rconn
```
Stebalien: Ideally, we'd mark his as a _fallback_ connection so we can stop using it when we get a real global connection.

reuse.go
```diff
@@ -0,0 +1,140 @@
...
+
+	// We don't have a connection that we can use for dialing.
+	// Dial a new connection from a random port.
+	var addr *net.UDPAddr
+	switch network {
+	case "udp4":
+		addr, err = net.ResolveUDPAddr("udp4", "0.0.0.0:0")
+	case "udp6":
+		addr, err = net.ResolveUDPAddr("udp6", "[::]:0")
+	}
```
Stebalien:
```go
net.UDPAddr{
  IP: net.IPv6zero // or net.IPv4zero,
  Port: 0,
  Network: "udpN",
}
```
?

reuse.go
Stebalien: We need to check when incrementing to see if the count dropped below 0.

(really, closing should set the count to -max).

reuse.go
```diff
@@ -0,0 +1,140 @@
...
+}
+
+func (c *reuseConn) IncreaseCount() { atomic.AddInt32(&c.refCount, 1) }
+func (c *reuseConn) DecreaseCount() { atomic.AddInt32(&c.refCount, -1) }
+func (c *reuseConn) GetCount() int  { return int(atomic.LoadInt32(&c.refCount)) }
+
+type reuse struct {
+	mutex sync.Mutex
+
+	unicast map[string] /* IP.String() */ map[int] /* port */ *reuseConn
```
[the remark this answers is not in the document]
Stebalien: other nit: do we really want to map port to conn or just `map[*reuseConn]struct{}`?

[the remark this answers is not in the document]
marten-seemann: It needs to be a map of port to conn, if we ever want to reuse a connection that we dialed on for listening. I admit, this is a rather rare case, but it doesn't cost us anything.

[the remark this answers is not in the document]
marten-seemann: I'm not sure I understand what you mean by `string(IP)`.

reuse.go
marten-seemann: Right, I copied that code from #52. Not sure where to close it, but at least we're using the same one within one "reuse" instance.
It should also make it a bit more explicit that this is only available on Linux.

reuse.go
marten-seemann: There's no `Network` in the `net.UDPAddr`, but I guess just `&net.UDPAddr{IP: net.IPv6zero, Port: 0}` should be fine.

reuse.go
marten-seemann: Right, we need to do this while holding the mutex. Fixed.

reuse.go
marten-seemann: I'm not sure I understand. `Dial` will use new connections if they're available, so this effective is already a fallback, isn't it?

reuse.go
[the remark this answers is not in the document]
Stebalien: > https://github.com/libp2p/go-yamux/pull/6#pullrequestreview-270734844

:+1:

> I'm not sure I understand what you mean by string(IP).

That is, we can convert the raw bytes directly to a string rather than formatting as "xxx.xxx.xxx.xxx"). It would make lookups zero-allocation (`mymap[string(someByteArray)]` doesn't allocate).

reuse.go
Stebalien: Events:

1. We dial, creating the fallback.
2. We listen on a specific port.
3. We dial again.

At step 3, we may use the connection from step 1 or step 3. Ideally, we'd use the connection from step 3.

transport.go
```diff
@@ -113,25 +123,35 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 	default:
 	}
 	if remotePubKey == nil {
+		pconn.DecreaseCount()
 		return nil, errors.New("go-libp2p-quic-transport BUG: expected remote pub key to be set")
 	}
+	go func() {
+		<-sess.Context().Done()
+		pconn.DecreaseCount()
+	}()
```
Stebalien: We can leave this for now but we may want to simply wrap close in the future. Leaving an extra goroutine open can get expensive.

reuse.go
Stebalien: We're racing against a separate (theoretical) garbage collection event. I guess we can have that hold the lock but it may be better to just _poison_ the connection such that `IncreaseCount` returns an error (and we try again).

reuse.go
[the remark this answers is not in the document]
marten-seemann: That's great! I wasn't aware that `net.IP` is just a `[]byte`.

reuse.go
marten-seemann: I guess you mean that ideally, we'd use the connection from step 2.

I'm going to merge this PR now, so we don't have to go through another round of review of this one. This also touches on reusing dialed connections for listening.

reuse.go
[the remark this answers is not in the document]
marten-seemann: Great in theory, that is. It looks like there's no obvious way to normalize IP addresses, and it's failing in our tests. 

I managed to reproduce the issue locally, the IP is 192.168.46.226.
The kernel resolves this to: `net.IP{0xc0, 0xa8, 0x2e, 0xe2}`.
After listening, the local address is: `net.IP{0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0xff, 0xff, 0xc0, 0xa8, 0x2e, 0xe2}`

Normalizing this looks straightforward in this case, however, I'm a bit wary of pitfalls when using IPv6 addresses. Map lookups will hardly be the most expensive operation when starting a new QUIC listener / dialer, so I think we'll be fine for now with leaving this as a `net.IP.String()`.

transport.go
marten-seemann: Not every connection is closed by calling `Close`. Only the peer that's actively closing the connection calls `Close`. The other peer will just receive a connection failure. Or the connection might time out, in which case nobody calls `Close` at all.

reuse.go
Stebalien: (yes, sorry)

PR #64, opened 2019-06-23
Fixed building

xaionaro:
github.com/lucas-clemente/quic-go broke an API:
* OpenStreamSync, AcceptStream, Accept now requires context.Context
* CloseWithError now requies string (not error)
* AcceptCookie was renamed to AcceptToken

```
 ../../libp2p/go-libp2p-quic-transport/conn.go:39:36: not enough arguments in call to c.sess.OpenStreamSync
         have ()
         want (context.Context)
 ../../libp2p/go-libp2p-quic-transport/conn.go:45:34: not enough arguments in call to c.sess.AcceptStream
         have ()
         want (context.Context)
 ../../libp2p/go-libp2p-quic-transport/listener.go:63:37: not enough arguments in call to l.quicListener.Accept
         have ()
         want (context.Context)
 ../../libp2p/go-libp2p-quic-transport/listener.go:69:24: cannot use err (type error) as type string in argument to sess.CloseWithError
 ../../libp2p/go-libp2p-quic-transport/transport.go:27:2: unknown field 'AcceptCookie' in struct literal of type quic.Config
```

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -1,6 +1,8 @@
 package libp2pquic
 
 import (
+	"context"
+
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/mux"
 	"github.com/libp2p/go-libp2p-core/peer"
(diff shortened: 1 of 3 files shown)


xaionaro:
The problem was I did not use go modules, so got too recent version of `github.com/lucas-clemente/quic-go`

Issue #60, opened 2019-05-01
QUIC transport constructor fails with ed25519 keys


vyzo:
Trying to bring up some relays using ed25519 keys, and I run into this:
```
May 01 18:14:27 ip-172-31-47-2 relay.sh[13132]: 2019/05/01 18:14:27 transport constructor github.com/libp2p/go-libp2p-quic-transport.NewTransport failed: unsupported key type for TLS
```

vyzo:
So basically we can't use QUIC with ed25519 keys.

Stebalien:
We need to switch to the new TLS handshake, IIRC.

vyzo:
Is there a quic(k) remedy for this so that we can use it now?

Stebalien:
Not really.

xaionaro:
I've just added a PR: https://github.com/libp2p/go-libp2p-quic-transport/pull/65

But it requires Go1.13

Kubuxu:
Yeah, we need to wait for Go1.13 as they've just introduced ed25519 support in TLS and included ed25519 in standard library.

marten-seemann:
We should probably rather reuse the code from the TLS package (see #18). In addition to code reuse, this will allow us to use all key types. Unfortunately, since QUIC doesn't have a separate `net.Conn` per connection, this is not trivial though.

xaionaro:
> We should probably rather reuse the code from the TLS package (see #18). In addition to code reuse, this will allow us to use all key types.

I've added a comment in the PR: https://github.com/libp2p/go-libp2p-quic-transport/pull/65#issuecomment-507534657

PR #65, opened 2019-07-01
Added support of ED25519 keys

xaionaro:
Works only with Go>=1.13

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -242,4 +242,26 @@ var _ = Describe("Connection", func() {
 		Eventually(done, 5*time.Second).Should(Receive())
 		Eventually(done, 5*time.Second).Should(Receive())
 	})
+
+	It("dials to ed25519 server", func() {
+		// Generate ED25519 credentials
+		serverKey2, _, err := ic.GenerateEd25519Key(rand.Reader)
+		Expect(err).ToNot(HaveOccurred())
+		serverID2, err := peer.IDFromPrivateKey(serverKey2)
+		Expect(err).ToNot(HaveOccurred())
+
+		// Create ED25519 server
+		serverTransport2, err := NewTransport(serverKey2)
+		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


xaionaro:
The test fails because of Go1.12. It's required to use Go1.13

marten-seemann:
As noted in https://github.com/libp2p/go-libp2p-quic-transport/issues/60#issuecomment-507524509, the better way to add ed25519 support is by reusing the code from the TLS package.

xaionaro:
@marten-seemann if you mean that this PR should be rejected because there's another task to reuse code from TLS package then I humbly disagree:

This tasks are two separate tasks which are not blocking each other. We can add support of ED25519 to the current code and while another task in someday after that we can declare current code obsolete and replace it by a new one (which will reuse the TLS package code).

This PR just improves the old code and allows people like me to use UDP with ED25519 right now (or right after official release of Go1.13). That's all.

marten-seemann:
@xaionaro We can't merge this PR at this time, since Go 1.13 is not released anyway. So it will be blocked for at least another 6 weeks. Furthermore, even after the release of Go 1.13 we have to worry about backwards compatibility: Nodes that haven't updated to a libp2p release compiled with Go 1.13 won't be able to complete handshake.

xaionaro:
> @xaionaro We can't merge this PR at this time, since Go 1.13 is not released anyway. So it will be blocked for at least another 6 weeks.

Yes, I understand that. But it at least has a more-less specific date. And it's not that far away from now.

> Furthermore, even after the release of Go 1.13 we have to worry about backwards compatibility: Nodes that haven't updated to a libp2p release compiled with Go 1.13 won't be able to complete handshake.

As I see IPFS already depends on Go1.12 which is currently the latest version. So in my understanding when Go1.13 will be released then IPFS will migrate on it anyway.

marten-seemann:
#67 is the way to go.

PR #66, opened 2019-07-09
resolve unspecified addresses

jbenet:
this fixes a tricky bug i found while trying to connect quic nodes
to each other in my local machine. They weren't able to connect to
each other because they weren't finding each other via the dht --
they would fail to advertise addresses to each other. (this bug
caused advertising of addrs via the identify protocol to fail).

unspecified (0.0.0.0) addresses are not meant to be part of
connections. libp2p assumes that conns have specified addresses.
0.0.0.0 addrs from listeners need to be resolved when assigned to
a connection.

Im not sure how transports like TCP achieve this (is this in manet?
default from golang/net package?), but there is a difference between
QUIC and TCP. For now, im using the addrutil function, but this is
probably not the right thing to do.

### DIFF
--- a/quic_multiaddr.go
+++ b/quic_multiaddr.go
@@ -1,8 +1,10 @@
 package libp2pquic
 
 import (
+	"errors"
 	"net"
 
+	addrutil "github.com/libp2p/go-addr-util"
 	ma "github.com/multiformats/go-multiaddr"
 	manet "github.com/multiformats/go-multiaddr-net"
 )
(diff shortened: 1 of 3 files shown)


quic_multiaddr.go
```diff
@@ -28,3 +30,21 @@ func toQuicMultiaddr(na net.Addr) (ma.Multiaddr, error) {
...
+		return nil, err
+	}
+	rma, err := addrutil.ResolveUnspecifiedAddresses([]ma.Multiaddr{qma}, nil)
+	if err != nil {
+		return nil, err
+	}
+	if len(rma) < 1 {
+		return nil, errors.New("failed to resolve multiaddr")
+	}
+	return rma[0], nil
```
jbenet: this is not the right way to do this, because the first addr here is probably not the right one. We should be able to grab the resolved socket addr from the os.

jbenet: (i believe this is strictly better than what it currently does though)

jbenet:
The real problem here is that `quic-go` is not returning the right address in the connection -- it should return the address assigned by the OS (probably the local address associated w/ the packets coming in, as described in https://github.com/lucas-clemente/quic-go/issues/1736, and in the POC linked there, not the address of the listener)

marten-seemann:
Yeah, that’s a really nasty issue due to the way Go exposes this information. Still haven’t figured out a good way to solve it.
The PoC only works when using a raw UDP conn, not when passing in a net.PacketConn. That’s a use case we can’t afford to lose.

Stebalien:
So... I think the issue @jbenet was talking about was fixed in: https://github.com/libp2p/go-libp2p/commit/5da711dc6c1ffb9f79258b53758072a62e64ddd9. I'm not sure why he was seeing this behavior.

Jorropo:
@jbenet like @Stebalien says I don't know how it doesn't work for you, for libp2p/go-utp-transport#5 returning an unspecified address works just well and get resolved without any issue.

Also I'm building the maddr with [manet.FromNetAddr](https://github.com/libp2p/go-utp-transport/blob/26c335a929bf58a9de25fd641b77ba720d799552/listener.go#L27) just to get the port right and [encapsulating](https://github.com/libp2p/go-utp-transport/blob/26c335a929bf58a9de25fd641b77ba720d799552/listener.go#L35) it with the non thinWaist part.

I know that counter intuitive but returning 0.0.0.0 as listener address works and get transformed for each interface (you just need to get the real port assigned by `:0`), also its not possible to return multiple maddr from one listener so it would be currently impossibe to return the multiple maddr of what 0.0.0.0 represent.

Stebalien:
I finally found the bug! @jbenet was using the stabilize branch in go-libp2p which didn't announce localhost addresses when the local end of the connection isn't localhost.

PR #67, opened 2019-07-29
use the handshake logic from go-libp2p-tls

### LINKED ISSUE #3: Handshake Design
## Design Rationale

This is based on the discussion in https://github.com/libp2p/specs/issues/37.

This document describes how two libp2p peers authenticate each other over QUIC. Note that we can use the same handshake not only for QUIC, but for any connection that uses TLS. It works with TLS 1.2 as well as TLS 1.3.

At initialisation, each peer generates a certificate chain consisting of two certificates:

1. a root certificate from the [libp2p private key](https://godoc.org/github.com/ipfs/go-libp2p-crypto#PrivKey). This is the only step that involves the private key, and consequently *could* be used to run a node that doesn't keep its private key in memory (if we ever decide to remove the getter method from the [`ConnSecurity` interface](https://github.com/libp2p/go-libp2p-net/blob/feat/refactor/interface.go#L42-L54))
2. an ephemeral key, that is supposed to be used for a limited time (currently: for the lifetime of the QUIC transport struct). This key is used to generate another certificate, which is signed by the root certificate.

Remember that in IPFS, the peer ID is derived from the key pair, and a node's identity can be verified by checking that their peer ID [matches](https://godoc.org/github.com/ipfs/go-libp2p-peer#ID.MatchesPublicKey) their public key.

Now each node has a certificate chain consisting of two certificates: 1. the root certificate, tied to the node's private key, and 2. an ephemeral certificate. We use TLS client authentication, so during the handshake each node will learn about the other's certificate chain. 

A client verifies a server by setting the `  tls.Config.VerifyPeerCertificate` callback, which is executed *during* the handshake (as soon as the server's cert chain is received). It does so by

1. parsing the certificate chain presented by the server and checking that it actually consists of 2 certificates
2. verifying that the public key from the root certificate matches the peer ID it intended to dial.

The server request TLS client authentication, but it doesn't run set the `VerifyPeerCertificate` callback. As soon as the TLS handshake completes, it

1. parsing the certificate chain presented by the client and checks that it actually consists of 2 certificates
2. determines the peer ID from the root certificate.

## Open Issues

- When the client sends an invalid cert chain (e.g. a chain that only contains one certificate), the server will not abort the handshake (since it only parses the cert chain after the handshake completes), but close the connection immediately after accepting. For the client this will look like the server first accepted the connection, and then closed it shortly afterwards. If we deem this unacceptable, we'd have to parse the cert chain twice (once using the `VerifyPeerCertificate` callback, and once after the handshake completes).

-  Do we need to version this? For example, if we ever want to change the length of the certificate chain to 3, this handshake would fail. We can't really negotiate a handshake protocol here, since authentication happens so early during connection establishment.

### LINKED ISSUE #18: reuse code from go-libp2p-tls

### LINKED ISSUE #60: QUIC transport constructor fails with ed25519 keys

marten-seemann:
Fixes #3. Fixes #18. Fixes #60.
CI is failing because quic-go dependency is not updated yet.

Furthermore, this depends on a new quic-go release, which includes https://github.com/lucas-clemente/quic-go/pull/2024. We will also need to increase the QUIC version number, since the old and the new handshake are incompatible. This makes sense anyway, since a new QUIC draft version was released recently. By using a new QUIC version number, peers using the old and the new handshake won't be able to connect to each other (using QUIC), so we won't get any handshake failures from that.

This PR implements the easiest possible way to reuse the handshake logic from go-libp2p-tls, which is slightly more computationally expensive for the server: Since QUIC multiplexes connections, we don't have the clear order that we have in TCP: 1. accept TCP connection, 2. run TLS handshake on that connection.
In QUIC, the first time we learn about a new connection attempt is when the `tls.Config.GetConfigForClient` callback is called. We use this callback to return a `tls.Config` that verifies the peer's certificate chain. At some point later, quic-go will return a fully establish QUIC connection (if the handshake succeeded). However, we have no way of associating this connection with the `tls.Config` we returned earlier (keep in mind that the server could be running multiple handshakes concurrently), so we need to parse the certificate chain a second time to determine the peer's identity.


transport.go
```diff
@@ -145,6 +120,14 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	if err != nil {
 		return nil, err
 	}
+
+	// Should be ready by this point, don't block.
+	var remotePubKey ic.PubKey
+	select {
+	case remotePubKey = <-keyCh:
+	default:
```
Stebalien: We should return an error in this case.

marten-seemann: That applies to [the equivalent code in go-libp2p-tls](https://github.com/libp2p/go-libp2p-tls/blob/cc53c4f8deb88e5de59405fb70ef27c93ba57ae1/transport.go#L102-L107) as well. I'll prepare a PR.

Stebalien: Same as in go-libp2p-tls, we probably also need to check if the channel is closed. Also, this check is backwards (return in the default case).

marten-seemann:
I just released a new quic-go version, which bumps the QUIC version number. See PR #68.
We should merge these two PRs at the same time, and then cut a new go-libp2p-quic-transport release.

Stebalien:
(lgtm once the tests pass)

Issue #57, opened 2019-04-10
goroutine build up stuck in dialing through quic


vyzo:
We are seeing a slow goroutine build up in our relays, stuck in dialing through quic:
```
    165407   runtime.gopark
             runtime.goparkunlock
             sync.runtime_notifyListWait
             sync.(*Cond).Wait
             github.com/lucas-clemente/quic-go.(*outgoingBidiStreamsMap).OpenStreamSync
             github.com/lucas-clemente/quic-go.(*streamsMap).OpenStreamSync
             github.com/lucas-clemente/quic-go.(*session).OpenStreamSync
             github.com/libp2p/go-libp2p-quic-transport.(*conn).OpenStream
             github.com/libp2p/go-libp2p-swarm.(*Conn).NewStream
             github.com/libp2p/go-libp2p-swarm.(*Swarm).NewStream
             github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).newStream
             github.com/libp2p/go-libp2p/p2p/host/basic.(*BasicHost).NewStream
             github.com/libp2p/go-libp2p-circuit.(*Relay).DialPeer
             github.com/libp2p/go-libp2p-circuit.(*Relay).Dial
             github.com/libp2p/go-libp2p-circuit.(*RelayTransport).Dial
             github.com/libp2p/go-libp2p-swarm.(*Swarm).dialAddr
             github.com/libp2p/go-libp2p-swarm.(*dialLimiter).executeDial
```

The reason for _dialing_ these peers in the first place is yet unknown, but the build up is more alarming and indicative of a bug.
The daemon is running with the latest `go-libpp2p-quic-transport` relase, v0.0.3

marten-seemann:
I'm not sure I understand the issue here. You're trying to open a streams, and at some point that blocks because you're only allowed to open a limited number of streams.

Stebalien:
How many streams do we allow by default? We should probably _fail_ when we hit such a limit instead of blocking (like "too many file descriptors").

Stebalien:
How many streams do we allow by default? We should probably _fail_ when we hit such a limit instead of blocking (like "too many file descriptors").

marten-seemann:
We allow 1000 streams: https://github.com/libp2p/go-libp2p-quic-transport/blob/2c4db4719d8a5d5117b16a81fde8a0ac0daa3f08/transport.go#L21-L22
[...]

PR #68, opened 2019-08-05
update quic-go to v0.12.0 (supporting QUIC draft-22)

marten-seemann:
This PR updates quic-go to the latest release. This release increases the QUIC version number to draft-22. This gives us the opportunity to change our handshake logic, so we should merge this PR at the same time as #67.


conn.go
```diff
@@ -36,13 +38,13 @@ func (c *conn) IsClosed() bool {
 
 // OpenStream creates a new stream.
 func (c *conn) OpenStream() (mux.MuxedStream, error) {
-	qstr, err := c.sess.OpenStreamSync()
+	qstr, err := c.sess.OpenStreamSync(context.Background())
```
Stebalien: We should probably set a timeout. 5s?

(ideally, the context would be passed through but that's better than nothing)

marten-seemann: That might be a fix for #57. Let's do that in a separate PR.

PR #73, opened 2019-08-07
implement a garbage-collector for unused reuse connections

### LINKED ISSUE #70: close unused connections
See the second half of https://github.com/libp2p/go-libp2p-quic-transport/pull/63#issue-284337030.

marten-seemann:
Fixes #70.

This PR implements the sweeper as suggested in #63.

The sweeper runs every 30 seconds and removes connections that have been unused for more than 10 seconds. In order to not leak go routines, it stops itself once a `reuse` doesn't have any more active connections, and is restarted as soon as a connection is added.

### DIFF
--- a/reuse.go
+++ b/reuse.go
@@ -132,20 +132,20 @@ func (r *reuse) Listen(network string, laddr *net.UDPAddr) (*reuseConn, error) {
 	defer r.mutex.Unlock()
 
 	// Deal with listen on a global address
-	if laddr.IP.IsUnspecified() {
+	if localAddr.IP.IsUnspecified() {
 		// The kernel already checked that the laddr is not already listen
 		// so we need not check here (when we create ListenUDP).
-		r.global[laddr.Port] = rconn
+		r.global[localAddr.Port] = rconn
 		return rconn, err
 	}
 
 	// Deal with listen on a unicast address
 	if _, ok := r.unicast[localAddr.IP.String()]; !ok {
-		r.unicast[laddr.IP.String()] = make(map[int]*reuseConn)
+		r.unicast[localAddr.IP.String()] = make(map[int]*reuseConn)
 	}
 
 	// The kernel already checked that the laddr is not already listen
 	// so we need not check here (when we create ListenUDP).
-	r.unicast[laddr.IP.String()][localAddr.Port] = rconn
+	r.unicast[localAddr.IP.String()][localAddr.Port] = rconn
 	return rconn, err
 }
--- a/reuse.go
+++ b/reuse.go
@@ -84,12 +84,14 @@ func (r *reuse) runGarbageCollector() {
 		r.mutex.Lock()
 		for key, conn := range r.global {
 			if conn.ShouldGarbageCollect(now) {
+				conn.Close()
 				delete(r.global, key)
 			}
 		}
 		for ukey, conns := range r.unicast {
 			for key, conn := range conns {
 				if conn.ShouldGarbageCollect(now) {
+					conn.Close()
 					delete(conns, key)
 				}
 			}
--- a/reuse.go
+++ b/reuse.go
@@ -3,26 +3,55 @@ package libp2pquic
 import (
 	"net"
 	"sync"
-	"sync/atomic"
+	"time"
 
 	"github.com/vishvananda/netlink"
 )
 
+// Constants. Defined as variables to simplify testing.
+var (
+	garbageCollectInterval = 30 * time.Second
+	maxUnusedDuration      = 10 * time.Second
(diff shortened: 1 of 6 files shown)


reuse.go
```diff
@@ -46,6 +75,51 @@ func newReuse() (*reuse, error) {
 	}, nil
 }
 
+func (r *reuse) runGarbageCollector() {
+	ticker := time.NewTicker(garbageCollectInterval)
+	for {
```
Stebalien: nit:

```
for range ticker.C {}
```

reuse.go
```diff
@@ -46,6 +75,51 @@ func newReuse() (*reuse, error) {
 	}, nil
 }
 
+func (r *reuse) runGarbageCollector() {
+	ticker := time.NewTicker(garbageCollectInterval)
```
Stebalien: Defer `ticker.Stop()`.

reuse.go
```diff
@@ -46,6 +75,51 @@ func newReuse() (*reuse, error) {
...
+			for key, conn := range conns {
+				if conn.ShouldGarbageCollect(now) {
+					delete(conns, key)
+				}
+			}
+			if len(conns) == 0 {
+				delete(r.unicast, ukey)
+			}
+		}
+
```
Stebalien: Don't we need to close sockets somewhere in here?

marten-seemann: Right, that's the whole point of this PR 😬

reuse.go
```diff
@@ -46,6 +75,53 @@ func newReuse() (*reuse, error) {
 	}, nil
 }
 
+func (r *reuse) runGarbageCollector() {
+	ticker := time.NewTicker(garbageCollectInterval)
+	defer ticker.Stop()
+
+	for range ticker.C {
```
Stebalien: uber nit: `for now := range ticker.C {`. We can skip the second call to `now`.

reuse.go
```diff
@@ -3,27 +3,56 @@ package libp2pquic
...
 }
 
 func newReuseConn(conn net.PacketConn) *reuseConn {
 	return &reuseConn{PacketConn: conn}
 }
 
-func (c *reuseConn) IncreaseCount() { atomic.AddInt32(&c.refCount, 1) }
-func (c *reuseConn) DecreaseCount() { atomic.AddInt32(&c.refCount, -1) }
-func (c *reuseConn) GetCount() int  { return int(atomic.LoadInt32(&c.refCount)) }
+func (c *reuseConn) IncreaseCount() {
```
Stebalien: nit: should probably be private.

reuse.go
```diff
@@ -3,27 +3,56 @@ package libp2pquic
...
+func (c *reuseConn) DecreaseCount() {
+	c.mutex.Lock()
+	c.refCount--
+	if c.refCount == 0 {
+		c.unusedSince = time.Now()
+	}
+	c.mutex.Unlock()
+}
+
+func (c *reuseConn) ShouldGarbageCollect(now time.Time) bool {
```
Stebalien: nit: also should be private.

reuse.go
```diff
@@ -3,27 +3,56 @@ package libp2pquic
...
-func (c *reuseConn) DecreaseCount() { atomic.AddInt32(&c.refCount, -1) }
-func (c *reuseConn) GetCount() int  { return int(atomic.LoadInt32(&c.refCount)) }
+func (c *reuseConn) IncreaseCount() {
+	c.mutex.Lock()
+	c.refCount++
+	c.unusedSince = time.Time{}
+	c.mutex.Unlock()
+}
+
+func (c *reuseConn) DecreaseCount() {
```
Stebalien: also should be private.

reuse.go
marten-seemann: Not an uber nit, this saves one syscall per tick.

reuse.go
marten-seemann: I don't think it actually matters. We're never exporting a `reuseConn` from this package, so there's no way a user could call this function anyway.

reuse.go
marten-seemann: Fixed.

Stebalien: Depends on the platform. Go will use vDSO on Linux.

reuse.go
Stebalien: Hm. But can't the user type assert?

marten-seemann: Not really. What we're returning is (wrapped) QUIC connections, not UDP connections.

Stebalien: Ah, fair enough. I still think private functions are clearer but, well, we're not consistent about that anyways.

PR #87, opened 2019-12-04
update quic-go to v0.14.0

marten-seemann:


marten-seemann:
Yes, and it's bumping the version number, so this won't break any connections.

Stebalien:
> Yes, and it's bumping the version number, so this won't break any connections.

But will old nodes be able to talk to new nodes?

marten-seemann:
No, we don't have any backwards compatibility. This is the reason why it's still experimental.

Stebalien:
Got it.

I'd like to release _a_ stable QUIC version ASAP, even if it's not the final RFC. We need QUIC and I don't want to keep waiting on the IETF.

However, when we do this, we'll need to support it into the future (i.e., support multiple versions). Is there any reason not to pick draft 24?

marten-seemann:
> However, when we do this, we'll need to support it into the future (i.e., support multiple versions). Is there any reason not to pick draft 24?

draft-25 will be the draft version that's intended for deployment by the IETF working group. There are a couple of important changes between -24 and -25, and it would be hard to support both at the same time. I suggest we wait for -25 before making QUIC the default transport.

Stebalien:
Got it. Any idea when draft 25 likely to be finalized?

marten-seemann:
I assume it will ship before the next QUIC working group interim meeting end of January next year.

Stebalien:
Awesome!

Stebalien:
@marten-seemann tell me when you've merged this so I can bubble it to go-ipfs. I'd like to get it into the release.

PR #93, opened 2019-12-07
test: libp2p transport test suite

marten-seemann:
Taking over #13. Can't figure out how to commit to the original branch and continue that PR.


marten-seemann:
The tests are failing because we're running into the `StressTestTimeout` (60s).

Stebalien:
@marten-seemann it would be _really_ nice to get this passing.

liamsi:
I've updated to the latest test-suite and made the build pass again: https://github.com/liamsi/go-libp2p-quic-transport/tree/transport-test-updated

The tests fail locally but with other fails than `StressTestTimeout (60s)` (or at least I don't see it if this happens under the hood): https://github.com/liamsi/go-libp2p-quic-transport/runs/3039813604

Stebalien:
@liamsi pulled.

Stebalien:
Ok, after https://github.com/libp2p/go-libp2p-testing/pull/34, everything but windows is passing, and windows is failing consistently. This _looks_ like it might be a QUIC bug.

Stebalien:
Hm. I take that back? I'm getting:

>       stream_suite.go:241: timeout: no recent network activity

On ubuntu as well, but not as reliably.

Also, after a quite a few runs, I'm getting "open sockets", even though I'm cleaning them up. Clearly we're something is failing to garbage collect (could be the test?).

Stebalien:
> Also, after a quite a few runs, I'm getting "open sockets", even though I'm cleaning them up. Clearly we're something is failing to garbage collect (could be the test?).

Ok, that's just us hanging on to old sessions.

Stebalien:
"retired connections"

Issue #74, opened 2019-08-13
Private network support


Stebalien:
Currently, the QUIC transport is incompatible with private networks as it doesn't use the pre-shared key in any way.

We have a couple of options:

1. Implement the proposed TLS1.3 PSK extension: https://tools.ietf.org/html/draft-ietf-tls-tls13-cert-with-extern-psk-00. This allows mixing the pre-shared key with the result from the handshake.
2. Add the ability to layer on some kind of trivial packet-based encryption protocol. We can use the same crypto as the current private network library as long as we use a new nonce per packet.

marten-seemann:
> Implement the proposed TLS1.3 PSK extension: https://tools.ietf.org/html/draft-ietf-tls-tls13-cert-with-extern-psk-00. This allows mixing the pre-shared key with the result from the handshake.

We're in the unfortunate situation where the Go team effectively froze `crypto/tls` (with the exception of security-related bugfixes). It doesn't seem like they are planning to implement **any** new features.
quic-go already uses a fork of `crypto/tls`, however, I'd feel more confident in the security properties if we don't mess too much with the key schedule.

> Add the ability to layer on some kind of trivial packet-based encryption protocol. We can use the same crypto as the current private network library as long as we use a new nonce per packet.

This is something we could do outside of quic-go. Effectively, we'd encrypt every QUIC packet that's being sent out. We'd either have to define a header that specifies the nonce of the packet, or use part of the QUIC packet payload as a nonce (for example the last 16 bytes of the packet, and leave those unencrypted). Aside from the obvious downside of double encryption, this would run against our efforts to make our traffic resemble commonly used internet protocols though.

Stebalien:
I'm leaning towards just encrypting each packet with a format: `<nonce><enc-message><hmac>`. That's what we do in go-libp2p-pnet (well, minus the MAC; we're relying on the actual transport for authentication in that case...).

Stebalien:
Having some form of private network (symmetric key) support in QUIC is needed before we can switch over to it by default. What's the status here? Can we mix in a symmetric key or are we going to have to introduce a packet-based pnet implementation?

marten-seemann:
There's not been any progress in Go's standard library TLS implementation, which we'd need to mix in symmetric keys.

Is there a third option here? Can we do the encryption on a stream level?

Stebalien:
> Is there a third option here? Can we do the encryption on a stream level?

[...]

PR #111, opened 2020-02-20
accept a PSK in the transport constructor (and reject it)

marten-seemann:
Don't construct a QUIC transport if a PSK is provided, since QUIC doesn't provide private network support yet (see #74).

This PR mainly serves to demonstrate how we'd use the new `ipnet.PSK` for QUIC.

Tracking issue: https://github.com/libp2p/go-libp2p/issues/795.


transport.go
```diff
@@ -86,7 +89,11 @@ type transport struct {
 var _ tpt.Transport = &transport{}
 
 // NewTransport creates a new QUIC transport
-func NewTransport(key ic.PrivKey) (tpt.Transport, error) {
+func NewTransport(key ic.PrivKey, psk pnet.PSK) (tpt.Transport, error) {
+	if len(psk) > 0 {
+		log.Error("QUIC doesn't support private networks yet.")
```
Stebalien: Note: when we actually implement this, let's derive a new shared secret by hashing `"quic-pnet" + PSK`.

marten-seemann: @Stebalien I've proposed to use a HKDF for that in https://github.com/libp2p/specs/pull/250

PR #113, opened 2020-02-28
Respect mux.ErrReset

Wondertan:
See libp2p/go-libp2p#814 for details

### DIFF
--- a/stream.go
+++ b/stream.go
@@ -6,14 +6,36 @@ import (
 	quic "github.com/lucas-clemente/quic-go"
 )
 
+const (
+	reset quic.ErrorCode = 0
+)
+
 type stream struct {
 	quic.Stream
 }
 
-var _ mux.MuxedStream = &stream{}
+func (s *stream) Read(b []byte) (n int, err error) {
+	n, err = s.Stream.Read(b)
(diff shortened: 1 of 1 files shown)


stream.go
```diff
@@ -6,14 +6,36 @@ import (
 	quic "github.com/lucas-clemente/quic-go"
 )
 
+const (
+	reset quic.ErrorCode = 0
```
Wondertan: I kept 0 as it was before, but it is better to introduce other special code

Wondertan:
@marten-seemann got your point. Then it makes sense to start defining error codes in specs. What do you think?

P.S. 
I know that the changes I am introducing here are not that important, but they are derived from the will to bring consistency between Stream implementations at least in Golang and potentially in other languages, so I expect any kind of counter-proposal on changes if they are not satisfying.

marten-seemann:
> @marten-seemann got your point. Then it makes sense to start defining error codes in specs. What do you think?

I don't think that's necessary, we can leave them unspecified for now. Regarding your PR, I'd just remove the check for the value of the error code (just do the interface assertion).

Wondertan:
Done.

PR #119, opened 2020-03-10
implement a QUIC protector

marten-seemann:


Stebalien:
Given that:

1. TCP still works with private networks.
2. https://tools.ietf.org/html/rfc8773 is the _correct_ approach.

We're closing this version.

PR #122, opened 2020-03-14
use a stateless reset key derived from the private key

### LINKED ISSUE #121: set a stateless reset token key

marten-seemann:
Fixes #121.

When a QUIC endpoint loses state (e.g. due to a reboot or a crash), it won't be able to decrypt packets that arrive for connection that it previously handled. QUIC uses a stateless reset to allow the peer to quickly detect this condition (and not having to rely on a lengthy connection timeout).

From the outside, a QUIC stateless reset packet looks like any other (Short Header) QUIC packet. An endpoint recognizes a stateless reset by comparing the last 16 bytes of the encrypted packet to a stateless reset token communicated earlier during the connection.

This token can't be tied to any connection state. Instead, it needs to be derived from values that survive a crash of the node. It therefore makes sense to derive it from the node's private key.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -117,7 +117,7 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK, filters *filter.Filters) (tpt.Tr
 	if err != nil {
 		return nil, err
 	}
-	keyReader := hkdf.Expand(sha256.New, keyBytes, []byte("libp2p quic stateless reset key"))
+	keyReader := hkdf.New(sha256.New, keyBytes, nil, []byte("libp2p quic stateless reset key"))
 	config.StatelessResetKey = make([]byte, 32)
 	if _, err := io.ReadFull(keyReader, config.StatelessResetKey); err != nil {
 		return nil, err
--- a/transport.go
+++ b/transport.go
@@ -2,11 +2,11 @@ package libp2pquic
 
 import (
 	"context"
-	"crypto/sha256"
 	"errors"
 	"io"
 	"net"
 
+	"github.com/minio/sha256-simd"
 	"golang.org/x/crypto/hkdf"
 
 	logging "github.com/ipfs/go-log"
--- a/transport.go
+++ b/transport.go
@@ -2,8 +2,12 @@ package libp2pquic
 
 import (
 	"context"
+	"crypto/sha256"
 	"errors"
+	"io"
 	"net"
+
+	"golang.org/x/crypto/hkdf"
 
 	logging "github.com/ipfs/go-log"
 	ic "github.com/libp2p/go-libp2p-core/crypto"
@@ -85,6 +89,7 @@ type transport struct {
 	localPeer   peer.ID
 	identity    *p2ptls.Identity
 	connManager *connManager
+	config      *quic.Config
 }
 
 var _ tpt.Transport = &transport{}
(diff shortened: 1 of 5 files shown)


transport.go
```diff
@@ -2,9 +2,13 @@ package libp2pquic
 
 import (
 	"context"
+	"crypto/sha256"
```
Stebalien: We usually use https://github.com/minio/sha256-simd for performance.

transport.go
```diff
@@ -106,12 +111,23 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK) (tpt.Transport, error) {
 	if err != nil {
 		return nil, err
 	}
+	config := quicConfig.Clone()
+	keyBytes, err := key.Raw()
+	if err != nil {
+		return nil, err
+	}
+	keyReader := hkdf.Expand(sha256.New, keyBytes, []byte("libp2p quic stateless reset key"))
```
Stebalien: Let's define the info as a constant.

Stebalien: Also, I don't think we can do this as the key may not be _uniformly_ random. I think we need the "extract" step (or need to call `New`)

I wonder if we should just _define_ some form of "DeriveSecretKey(info)` function on our secret key interface. That way, we have a standard way to do this.

marten-seemann: Sounds good. Let me prepare a PR.

transport.go
marten-seemann: I don't think performance is critical here, since the only time we calculate a hash is when `NewTransport` is called. I'm ok with changing it to keep things consistent though.

This will be obsoleted with https://github.com/libp2p/go-libp2p-core/pull/131 anyway.

transport.go
marten-seemann: @Stebalien Can we move forward with this PR without resolving https://github.com/libp2p/go-libp2p-core/pull/131 first? For the stateless reset token, the only thing that matter is that the peer has **any** was of deriving a key. No interoperability required.

Stebalien: You're right.

Stebalien: Note: if we do that, you'll need to update this patch to call `hkdf.New` instead of `hkdf.Expand`.

marten-seemann: @Stebalien Done.

transport.go
```diff
@@ -107,12 +112,23 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK, filters *filter.Filters) (tpt.Tr
 	if err != nil {
 		return nil, err
 	}
+	config := quicConfig.Clone()
+	keyBytes, err := key.Raw()
+	if err != nil {
+		return nil, err
+	}
+	keyReader := hkdf.New(sha256.New, keyBytes, nil, []byte("libp2p quic stateless reset key"))
```
Stebalien: 1. Let's make the "info" a variable/constant (private).
2. Let's add a _test_ to make sure we don't accidentally change it. That will cause silent breakage that won't be caught except in interop tests (which we need anyways but still).

marten-seemann: 1. Done.
2. I'm not convinced that's necessary. Stateless resets are an optimization, and the worst case behavior in the absence of one is a connection timeout.

Stebalien: That's even worse. That's a performance bug that we won't catch in _any_ tests because it'll "just work" even if we break it.

PR #125, opened 2020-03-15
add support for multiaddr filtering

### LINKED ISSUE #124: Filter inbound connections
Transports should apply the "address filters", either directly on all incoming packets or, at least, on the initial handshake packets. Applying them on all incoming packets could be a performance hit at the moment as the filter is _very_ unoptimized.

The transport can get a copy of the filter set by accepting a `*github.com/libp2p/go-maddr-filter.Filters` in the constructor.

marten-seemann:
Fixes #124.

When multiplexing incoming and outgoing connections on the same packet conn, quic-go expects the packet conns to be the same struct. Therefore, filters are applied for incoming and outgoing connections.

### DIFF
--- a/filtered_conn.go
+++ b/filtered_conn.go
@@ -1,0 +1,34 @@
+package libp2pquic
+
+import (
+	"net"
+
+	filter "github.com/libp2p/go-maddr-filter"
+)
+
+type filteredConn struct {
+	net.PacketConn
+
+	filters *filter.Filters
+}
+
+func newFilteredConn(c net.PacketConn, filters *filter.Filters) net.PacketConn {
+	return &filteredConn{PacketConn: c, filters: filters}
+}
+
+func (c *filteredConn) ReadFrom(b []byte) (n int, addr net.Addr, rerr error) {
+	for {
+		n, addr, rerr = c.PacketConn.ReadFrom(b)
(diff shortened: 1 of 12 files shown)


filtered_conn.go
```diff
@@ -0,0 +1,34 @@
...
+func newFilteredConn(c net.PacketConn, filters *filter.Filters) net.PacketConn {
+	return &filteredConn{PacketConn: c, filters: filters}
+}
+
+func (c *filteredConn) ReadFrom(b []byte) (n int, addr net.Addr, rerr error) {
+	for {
+		n, addr, rerr = c.PacketConn.ReadFrom(b)
+		// Short Header packet, see https://tools.ietf.org/html/draft-ietf-quic-invariants-07#section-4.2.
+		if n < 1 || b[0]&0x80 == 0 {
+			return
```
Stebalien: Can we ever read a partial packet and would that cause an issue?

marten-seemann: In principle, yes. If the UDP packet is longer than `b`, `ReadFrom` will only read the packet up to `len(b)`, the rest of the packet will be discarded. Since every QUIC packet is integrity protected, this doesn't lead to any problems, as the partial packet will simply end up undecryptable.
QUIC itself deals with the problem of fixed buffer sizes by communication a `max_packet_size` during the handshake, so the endpoint knows not to pack any packet that is larger than its peer's buffer.

PR #126, opened 2020-03-16
reject /dns addresses when dialing

### LINKED ISSUE #123: RemoteMultiaddr should be fully resolved
The addr returned by RemoteMultiaddr should be fully resolved, even if the user dials with a /dns address.

marten-seemann:
Fixes #123.

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -85,6 +85,15 @@ var _ = Describe("Connection", func() {
 		Expect(serverConn.RemotePublicKey().Equals(clientKey.GetPublic())).To(BeTrue())
 	})
 
+	It("rejects /dns addresses", func() {
+		clientTransport, err := NewTransport(clientKey, nil)
+		Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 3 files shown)


Stebalien:
Let's wait until we resolve the go-libp2p bug.

Really, this shouldn't be unnecessary, strictly speaking because, `CanDial` should return `false` when passed a dns addr. However, it currently returns `true` because [this now includes dns](https://github.com/multiformats/go-multiaddr-fmt/blob/7612701fafc8509b9d203b42715c5d704d15a886/patterns.go#L34), and QUIC is defined [here](https://github.com/multiformats/go-multiaddr-fmt/blob/7612701fafc8509b9d203b42715c5d704d15a886/patterns.go#L42).

marten-seemann:
`CanDial` is implemented by the QUIC transport. We can add the check there as well.

marten-seemann:
@Stebalien I just fixed the behavior of `CanDial` to return `false` for /dns addresses.

PR #127, opened 2020-03-16
use the resolved address for RemoteMultiaddr()

marten-seemann:
Not sure if this is really necessary, but it doesn't hurt to use the resolved address here.


Stebalien:
I agree. It's strictly more correct, but may not be necessary at the end of the day.

PR #131, opened 2020-04-03
fix: avoid dialing/listening on dns addresses

Stebalien:
See https://github.com/libp2p/go-libp2p/issues/841

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -190,9 +190,12 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	}, nil
 }
 
+// Don't use mafmt.QUIC as we don't want to dial DNS addresses. Just /ip{4,6}/udp/quic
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -190,9 +190,12 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	}, nil
 }
 
+// Don't use mafmt.QUIC as we don't want to dial DNS addresses. Just /ip{4,6}/udp/quic
```
Stebalien, line 193: @marten-seemann ^^

marten-seemann: 🤦‍♂️

Stebalien:
I completely forgot about that PR.

transport.go
Stebalien: Yeah, it's confusing. We may want to just go ahead and revert the change in mafmt (we changed it from IP only to IP+DNS). But it does make sense to allow DNS there, just not _here_.

PR #134, opened 2020-04-06
switch local route binding to use netroute

### LINKED ISSUE #130: Switch to go-netroute
@willscott just wrote a cross-platform package (github.com/libp2p/go-netroute) for getting the route to a target IP. We should replace the current dependency on the netlink package.

This will give proper reuse support on windows.

willscott:
Fixes #130.

This simplifies the logic in quic-transport on what listeners / source IPs to reuse.
Instead of the current netlink logic which is limited to linux, the [go-netroute](github.com/libp2p/go-netroute) library will expand support of querying OS routing tables to windows and mac as well. Removing the OS special casing appears to remove the need for the split between `reuse` and `reuseBase` in this package.


reuse.go
```diff
@@ -110,14 +111,34 @@ func (r *reuseBase) runGarbageCollector() {
...
+func (r *reuse) maybeStartGarbageCollector() {
 	if !r.garbageCollectorRunning {
 		r.garbageCollectorRunning = true
 		go r.runGarbageCollector()
 	}
 }
+func (r *reuse) Dial(network string, raddr *net.UDPAddr) (*reuseConn, error) {
+	ips := []net.IP{}
+	if router, err := netroute.New(); err == nil {
+		_, _, src, err := router.Route(raddr.IP)
```
marten-seemann, line 123: Do we need to handle `err != nil` here?

reuse.go
```diff
@@ -110,14 +111,34 @@ func (r *reuseBase) runGarbageCollector() {
...
 // must be called while holding the mutex
-func (r *reuseBase) maybeStartGarbageCollector() {
+func (r *reuse) maybeStartGarbageCollector() {
 	if !r.garbageCollectorRunning {
 		r.garbageCollectorRunning = true
 		go r.runGarbageCollector()
 	}
 }
+func (r *reuse) Dial(network string, raddr *net.UDPAddr) (*reuseConn, error) {
+	ips := []net.IP{}
```
marten-seemann, line 121: Why is this a slice? We'll only ever have (at most) a single entry, right?

willscott:
Rebase should be in place. can squash these commits before merging.

reuse.go
willscott: there isn't a great `nil` value for src - checking for nil error seems like a reasonable sanity check before validating that the source hasn't been left as all 0's.

PR #136, opened 2020-04-16
write qlogs to a temporary file first, then rename them when done

### LINKED ISSUE #137: Windows doesn't like our qlog file names
Specifically, it doesn't like `:` in the `RFC3339Nano` format. Looks like we have to roll out our own time stamp format.

marten-seemann:
In my tests on the staging server, I get a bunch of invalid `.qlog.gz` files. I assume the reason is that these are connections that are still active, and the log file is still being written.

It makes sense to use a different file name when writing the file. I chose the same naming scheme as vim, i.e. the temporary file for `log.qlog.gz` is `.log.qlog.gz.swp` (in the same directory). I'd be happy to change this to something different, if you have any suggestions.

This PR also uncovered that Windows doesn't like the file names of our qlog files (see #137). This went undetected because so far we didn't have a unit test that actually performed file system access.
This PR also fixes #137.


marten-seemann:
Lesson learned from countless CI failures: Developing on / for Windows is an absolute pain.

PR #152, opened 2020-05-18
gate QUIC connections via new ConnectionGater

aarshkshah1992:
For https://github.com/libp2p/go-libp2p/issues/931.

- This PR intercepts packets read from the UDP socket to determine if they should be passed upwards in the stack or gated right away. 
- It does so by calling the `InterceptAccept` method of the Connection Gater that is passed to the QUIC transport during initialization.

@raulk 
I am not sure if we can actually call `InterceptSecured` here as the security handshake seems to be part of making a QUIC connection and is handled completely by the QUIC library we use.


conn_test.go
```diff
@@ -11,17 +11,44 @@ import (
...
+
 	quicproxy "github.com/lucas-clemente/quic-go/integrationtests/tools/proxy"
 	ma "github.com/multiformats/go-multiaddr"
+	manet "github.com/multiformats/go-multiaddr-net"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+type localhostMockGater struct {
```
aarshkshah1992, line 28: @raulk I know this will make you cringe and add some entropy to thy life. But, I have added a TODO at https://github.com/libp2p/go-libp2p/issues/930 to replace this with the "Functional Gater" that we lan to implement. 

For all that you hold dear on this good Earth, rest assured, this shall be fixed.

filtered_conn.go
```diff
@@ -23,12 +44,15 @@ func (c *filteredConn) ReadFrom(b []byte) (n int, addr net.Addr, rerr error) {
...
-		if !c.filters.AddrBlocked(maddr) {
+
+		connAddrs := &connAddrs{lmAddr: c.lmAddr, rmAddr: rmAddr}
+
+		if c.connGater.InterceptAccept(connAddrs) {
 			return
 		}
 	}
-}
+}
```
raulk, line 58: go fmt missing.

transport.go
```diff
@@ -93,7 +94,7 @@ type transport struct {
 var _ tpt.Transport = &transport{}
 
 // NewTransport creates a new QUIC transport
-func NewTransport(key ic.PrivKey, psk pnet.PSK, filters *filter.Filters) (tpt.Transport, error) {
+func NewTransport(key ic.PrivKey, psk pnet.PSK, connGater connmgr.ConnectionGater) (tpt.Transport, error) {
```
raulk, line 97: Let's call this `gater` everywhere; there's no other gater in the system, and we don't expect there to be, so the `Connection` qualifier is redundant. (And it's idiomatic Go to shorten variable names the nearer they're used).

aarshkshah1992:
> Note: the `InterceptUpgraded` will be redundant with the swarm's call (we should remove it from there).

I am not sure why we need to call `InterceptUpgraded` here. The Swarm will call `InterceptUpgraded` for ALL transports. There's no special casing there. We don't call `InterceptUpgraded` in any of the other transports(upgrader) as well.

raulk:
@aarshkshah1992 It's a _separation of concerns_ argument. The component that's responsible for returning an upgraded connection is the transport, therefore it should be the one intercepting, in my view. However, OTOH the swarm receives _only_ upgraded connections from the transports, so it's safe to just leave the interception point there... No right or wrong here, disregard my comment about `InterceptUpgraded` and let's implement only `InterceptSecured` here.

raulk:
@aarshkshah1992 We should update the godocs in go-libp2p-core to make it explicit what's expected to be implemented by transport writers and what isn't...

transport.go
aarshkshah1992: This is done.

aarshkshah1992:
@raulk Have addressed your changes. Please take a look now.

conn_test.go
```diff
@@ -166,17 +193,12 @@ var _ = Describe("Connection", func() {
...
-		ipNet := net.IPNet{
-			IP:   net.IPv4(127, 0, 0, 1),
-			Mask: net.IPv4Mask(255, 255, 255, 255),
-		}
-		filters.AddFilter(ipNet, filter.ActionDeny)
 		testMA, err := ma.NewMultiaddr("/ip4/127.0.0.1/udp/1234/quic")
 		Expect(err).ToNot(HaveOccurred())
-		Expect(filters.AddrBlocked(testMA)).To(BeTrue())
+		cg := &localhostMockGater{}
+		Expect(cg.InterceptAccept(&connAddrs{rmAddr: testMA})).To(BeFalse())
```
raulk, line 203: This is testing the gater, rather than the code we've introduced.

conn_test.go
```diff
@@ -192,7 +214,7 @@ var _ = Describe("Connection", func() {
 
 		// now allow the address and make sure the connection goes through
 		clientTransport.(*transport).clientConfig.HandshakeTimeout = 2 * time.Second
-		filters.AddFilter(ipNet, filter.ActionAccept)
+		cg.allowAll = true
```
raulk, line 222: Aren't we missing a test that actually tests the interception?

conn_test.go
raulk: Well, it seemed to be the case before too. I guess a sanity check.

conn_test.go
```diff
@@ -192,7 +218,9 @@ var _ = Describe("Connection", func() {
 
 		// now allow the address and make sure the connection goes through
 		clientTransport.(*transport).clientConfig.HandshakeTimeout = 2 * time.Second
-		filters.AddFilter(ipNet, filter.ActionAccept)
+		cg.lk.Lock()
+		cg.allowAll = true
```
raulk, line 222: A test for `InterceptSecured` would be great (mostly for regression testing).

listener.go
```diff
@@ -3,6 +3,8 @@ package libp2pquic
 import (
 	"context"
 	"crypto/tls"
+	"fmt"
+	n "github.com/libp2p/go-libp2p-core/network"
```
raulk, line 7: import order

conn_test.go
aarshkshah1992: @raulk Yes, hence left it there.

filtered_conn.go
```diff
@@ -23,11 +44,14 @@ func (c *filteredConn) ReadFrom(b []byte) (n int, addr net.Addr, rerr error) {
...
 		if err != nil {
 			panic(err)
 		}
-		if !c.filters.AddrBlocked(maddr) {
+
+		connAddrs := &connAddrs{lmAddr: c.lmAddr, rmAddr: rmAddr}
+
+		if c.gater.InterceptAccept(connAddrs) {
 			return
 		}
```
raulk, line 56: This is interesting. I don't think it translates to the new model properly. Our `InterceptAccept` is supposed to be called once, when we decide to accept or reject an incoming connection. The previous implementation is calling the filters _on every single network packet_. Have a look through #124 -- there's some history there.

@marten-seemann -- do you recall why incoming filtering wasn't implemented [here instead](https://github.com/libp2p/go-libp2p-quic-transport/blob/master/reuse.go#L183)? We could call the `InterceptAccept` and `Close()` the UDP connection if rejected. Alternatively if we want to avoid the kernel notifying us new packets from the same source as new UDP connections, we could just "drain" the connection, which is effectively what we're doing now anyway.

listener.go
aarshkshah1992: This is fixed.

filtered_conn.go
Stebalien: `newReuseConn` is called once per UDP "connection". That is, once per address we're listening on.

In this case, we call `InterceptAccept` once per packet that starts with `0x80`. Those are packets with the QUIC "long" header. Usually, these packets are only sent for the initial handshake.

The alternative would be to add low-level hooks into the QUIC transport itself before QUIC completes the crypto handshake.

Stebalien:
Note: It's not actually possible to call `InterceptUpgraded` here as we have a `transport.CapableConn`, not a `network.Conn`.

conn_test.go
aarshkshah1992: This is done.

aarshkshah1992:
@raulk Have added a test for `InterceptSecured` as well. Please approve and merge if happy.

filtered_conn.go
raulk: Right, I had a senior moment and got confused by golang's use of the term "Connection" to refer to a UDP listener.

raulk:
> Note: It's not actually possible to call `InterceptUpgraded` here as we have a `transport.CapableConn`, not a `network.Conn`.

Correct. I knew there was a reason I had originally planned `InterceptUpgraded` to take a `transport.CapableConn`. But that wouldn't have worked anyway, because the calling component will, in the future, need to return a disconnect reason over a control stream, for which we need a `network.Conn` that's working with multistream.

conn_test.go
```diff
@@ -8,20 +8,54 @@ import (
...
+	manet "github.com/multiformats/go-multiaddr-net"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 )
 
+type mockGater struct {
+	lk                 sync.Mutex
+	allowAllAccepted   bool
+	disableSecuredPeer peer.ID
```
raulk: I've simplified these names @aarshkshah1992.

transport.go
```diff
@@ -178,6 +183,13 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 		pconn.DecreaseCount()
 		return nil, err
 	}
+
+	connaddrs := &connAddrs{lmAddr: localMultiaddr, rmAddr: remoteMultiaddr}
+	if t.gater != nil && !t.gater.InterceptSecured(n.DirOutbound, p, connaddrs) {
```
marten-seemann, line 188: We need to close the connection at this point. Otherwise the peer will never know that we closed the connection. As from quic-go's point of view the connection is fully established, there won't be any connection-level timeout here either.

PR #156, opened 2020-05-21
fix: call InterceptSecured earlier; drop conn.

raulk:
Fixes #155.

PR #157, opened 2020-05-21
close the connection when it is refused by InterceptSecured

### LINKED ISSUE #155: connection is not closed when rejected by InterceptSecured
See https://github.com/libp2p/go-libp2p-quic-transport/pull/152#pullrequestreview-415900175.

marten-seemann:
Minimally fixes #155. Closes #156.

This is the solution to #155 that's analogous to what we do for other transports: We first finish the cryptographic handshake (which verifies the peer's identity), and then afterwards close the connection when the connection gater doesn't like that peer's ID.

### DIFF
--- a/transport.go
+++ b/transport.go
@@ -180,13 +180,13 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 
 	localMultiaddr, err := toQuicMultiaddr(pconn.LocalAddr())
 	if err != nil {
-		pconn.DecreaseCount()
+		sess.CloseWithError(0, "")
 		return nil, err
 	}
 
 	connaddrs := &connAddrs{lmAddr: localMultiaddr, rmAddr: remoteMultiaddr}
 	if t.gater != nil && !t.gater.InterceptSecured(n.DirOutbound, p, connaddrs) {
-		pconn.DecreaseCount()
+		sess.CloseWithError(0, "")
 		return nil, fmt.Errorf("secured connection gated")
 	}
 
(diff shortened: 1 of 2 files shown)


transport.go
```diff
@@ -186,6 +186,7 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 
 	connaddrs := &connAddrs{lmAddr: localMultiaddr, rmAddr: remoteMultiaddr}
 	if t.gater != nil && !t.gater.InterceptSecured(n.DirOutbound, p, connaddrs) {
+		sess.CloseWithError(0, "")
```
marten-seemann, line 189: We *could* use a different error code and / or a reason string here. Might help debugging, but then this would be something QUIC-specific as afaik we don't have any close reasons for TCP-based transports.

marten-seemann:
Overhead in what sense?

raulk:
@marten-seemann creating state in QUIC that doesn't need to exist, which will be immediately available for GC? #156 rejects drops connection as soon as we authenticate the peer, and will likely perform better in terms of allocs and ns/op. We can benchmark it if you'd like.

marten-seemann:
quic-go creates almost all of the state needed for a connection when accepting it (not when the handshake completes), so I don't expect a big difference here. For sure, aborting early is almost always more efficient, but since this is not the common case anyway, I don't think it's worth optimizing too much for it. I'd rather be consistent with other transports here.

raulk:
@marten-seemann no other transport does this interception itself. They delegate on the Upgrader, which does it ASAP. We push the gater down to the transport precisely so that each transport can gate in the most efficient way, i.e. the goal is to _specialise_, not to _normalise_. I think the other PR achieves this in the most efficient and cleanest way possible.

raulk:
I don't think this is the right way to do this, @Stebalien. Please look at #156 and discussion.

Stebalien:
Sorry, I looked at the code instead of reading the discussion. I agree with @raulk here. We should reject the connection as _early_ as we can.

Stebalien:
Ok, I need to read the full discussion before commenting. Continuing in https://github.com/libp2p/go-libp2p-quic-transport/issues/155.

Stebalien:
Given that this PR is a strict bug fix while the other PR bundles an optimization, I'm going to merge this and let the debate continue.

PR #158, opened 2020-05-22
fix a potential nil-pointer panic.

raulk:
Cherry-picked from #156, while we sort the discussion out.


marten-seemann:
I don’t think this is necessary. The filtered conn is only created when we use a gater: 
https://github.com/libp2p/go-libp2p-quic-transport/blob/eda74852da7ddefc87b2c5db9ba34b2a7107f8cc/reuse.go#L28

Can you please revert this PR?

raulk:
Oh, right. The `filteredConn` is rather confusing.

PR #165, opened 2020-06-12
update quic-go to v0.16.2

marten-seemann:
This version supports both draft-28 and draft-29, with draft-29 as the preferred version (the version that's first used on dial attempts).


Stebalien:
Thank you :heart:.

PR #168, opened 2020-07-11
update quic-go to v0.17.3

marten-seemann:
This will be necessary to fix issue https://github.com/ipfs/go-ipfs/issues/7526.

Please note the release notes: https://github.com/lucas-clemente/quic-go/releases/tag/v0.17.3. I'm planning to set the `quic.RetireBugBackwardsCompatibilityMode` in go-ipfs.


Stebalien:
Nice! So this isn't a backwards incompatible change, it just includes an additional fix to work around peers that haven't upgraded, right?

marten-seemann:
Yes, the particular nature of this bug allows this. What this flag does it just disable the issuing of new connection IDs, so the bug won't be triggered.

PR #171, opened 2020-08-20
update quic-go to v0.18.0

marten-seemann:
We can now (finally!) build quic-go with Go 1.15 (Go 1.14 is also still supported, in case we need to downgrade for whatever reason).


transport.go
```diff
@@ -124,19 +124,17 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK, gater connmgr.ConnectionGater) (
 	if _, err := io.ReadFull(keyReader, config.StatelessResetKey); err != nil {
 		return nil, err
 	}
+	config.Tracer = tracer
```
raulk, line 127: Is it ok for this to be nil? i.e. when the `QLOGDIR` env var is not set?

qlog_test.go
```diff
@@ -16,18 +18,14 @@ type nopCloser struct{}
...
 
 var _ = Describe("qlogger", func() {
-	var origQlogDir string
+	var qlogDir string
 
 	BeforeEach(func() {
-		origQlogDir = qlogDir
-		d, err := ioutil.TempDir("", "libp2p-quic-transport-test")
+		var err error
+		qlogDir, err = ioutil.TempDir("", "libp2p-quic-transport-test")
```
raulk, line 25: We should probably do the equivalent of `t.Cleanup()` to delete the temp dir after the test is done.

transport.go
marten-seemann: Yes, that's ok. It's also ok if the callback returns a nil `io.WriteCloser` (that way you could sample only a fraction of the connections you're handling, for example).

qlog_test.go
marten-seemann: Done.

raulk (approved):

PR #173, opened 2020-08-25
use GitHub Actions to test interopability of releases

marten-seemann:
This GitHub Actions config:
1. builds a go-libp2p-quic-transport binary using the release version / Go version specified in `matrix.jsonc`
2. then runs interoperability tests between each and every of combination

This will allow us to have some amount of confidence that releasing a new version of this package doesn't break interoperability with older releases.

### DIFF
--- a/.github/workflows/interop.sh
+++ b/.github/workflows/interop.sh
@@ -1,0 +1,16 @@
+#!/bin/bash
+
+set -e 
+
+SERVER=$1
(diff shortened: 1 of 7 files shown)


marten-seemann:
I added a wrapper if the commit being built is before the change of the stream interface introduced in #174.

Stebalien (approved):

PR #187, opened 2020-11-20
use GitHub Actions instead of Travis

marten-seemann:
cc @mvdan

This was easier than I expected. It's little more than a copy-paste of the GitHub Actions configuration I use for quic-go.


Stebalien (approved):
This is beautiful.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,35 @@
...
+      - uses: actions/checkout@v2
+      - uses: actions/setup-go@v2
+        with: 
+          go-version: ${{ matrix.go }}
+      - run: go version
+      - name: Install test tools
+        run: |
+          go get golang.org/x/tools/cmd/cover
+          go get github.com/onsi/ginkgo/ginkgo
+          go get github.com/onsi/gomega
```
mvdan, line 21: assuming these tools are tracked in your go.mod, it would be best to use `go install`, as `go get` is increasingly geared towards updating/adding dependencies to go.mod.

.github/workflows/unit.yml
```diff
@@ -0,0 +1,35 @@
+on: [push, pull_request]
+
+jobs:
+  unit:
+    strategy:
+      matrix:
+        os: [ "ubuntu", "windows", "macos" ]
+        go: [ "1.14", "1.15" ]
```
mvdan, line 8: the same `.x` point as here: https://github.com/lucas-clemente/quic-go/pull/2888/files#r527797485

mvdan (approved):

.github/workflows/unit.yml
marten-seemann: I don't think that's necessary:
image

mvdan: Huh, that's very surprising. I'd personally still use `1.14.x` because it's very clear - Go's first 1.14 version is literally just called 1.14, not 1.14.0.

.github/workflows/unit.yml
marten-seemann: We need the `ginkgo` command line utility, so `go install` won't work. We don't need the cover tool and gomega though.

mvdan: Why wouldn't `go install` work? As long as these module deps are in go.mod, `go install` is better than `go get`, because `go get` is really meant for updating go.mod

marten-seemann: `go install` doesn't work, as we need the `ginkgo` binary. `go install github.com/onsi/ginkgo/ginkgo` works though.

marten-seemann: Is that better?

mvdan: Yes, that's better :) I did mean replacing `go get github.com/onsi/ginkgo/ginkgo` with `go install github.com/onsi/ginkgo/ginkgo`, perhaps I wasn't very clear.

PR #190, opened 2021-01-19
switch from gzip to zstd for qlog compression

marten-seemann:
When running a high-bandwidth connection, qlogging consumes about ~40% CPU, of which 1/4 is spent on JSON encoding an 3/4 is spent on running gzip. Switching from gzip to zstd reduces the time spent on compression to 1/2 of the time spent on JSON encoding. That's a speedup of 6x.

Changing the compression level doesn't actually have a large influence on the file size (in my tests I wasn't able to gain more than 10% by increasing it), so using the fastest compression makes sense here. The zst file will be roughly as large as the gzip file.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,6 +5,7 @@ go 1.14
 require (
 	github.com/golang/mock v1.4.4
 	github.com/ipfs/go-log v1.0.4
+	github.com/klauspost/compress v1.11.7
 	github.com/libp2p/go-libp2p-core v0.8.0
 	github.com/libp2p/go-libp2p-tls v0.1.3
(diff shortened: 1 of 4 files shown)


Stebalien (approved):
LGTM as long as this builds fine with CGO_ENABLED=0.

marten-seemann:
Why would it not? klauspost/compress/zstd is a pure Go implementation.

Stebalien:
It should be fine. I just want to make sure because other packages in that repo do use CGO, and sometimes the CGO dependency ends up leaking if people aren't careful.

marten-seemann:
Just confirmed that it actually builds with CGO_ENABLED=0. It does.

PR #193, opened 2021-02-11
compress qlogs when the QUIC connection is closed

### LINKED ISSUE #191: zstd memory consumption kills nodes
Running an IPFS node on a virtual server equipped with 2 GB of RAM leads to OOM within 2 - 5 minutes when qlog is running. The reason is that each `zstd.Encoder` consumes at least 8 MB of memory for an internal buffer, so just handling a little more than 200 QUIC connections simultaneously would consume all available memory.

I reported this to the zstd implementation in https://github.com/klauspost/compress/issues/316.

I hope that this can be fixed there, otherwise we'd probably have to log the uncompressed qlog to disk first, and do the compression in a post-processing step when the respective connection is closed. This would reduce the number of simultaneously allocated `zstd.Encoder`s.

### LINKED ISSUE #192: qlog compression is a memory hog
Compressing qlogs uses huge amounts of memory.
Screen Shot 2021-02-11 at 11 49 00

Running v0.10.0 (which uses gzip, not zstd), so this is unrelated to #191.

The problem is that we're running as many compression contexts in parallel as we have active QUIC connections. This number can easily reach hundreds.

marten-seemann:
Fixes #191. Fixes #192.

I hate that we to make this change, but I don't see any other solution. Compression algorithms keep a lot of internal state, so we shouldn't have too many of them running in parallel. This was already an issue when using gzip (see #192), which got exacerbated by using zstd.

The alternative is to use the file system as a buffer: We qlog into a temporary file `.<name>.qlog.swp`. Once the QUIC connection is closed, we compress that file to `<name>.qlog.zst`.

### DIFF
--- a/tracer.go
+++ b/tracer.go
@@ -68,6 +68,8 @@ func newQlogger(qlogDir string, role logging.Perspective, connID []byte) io.Writ
 }
 
 func (l *qlogger) Close() error {
+	defer os.Remove(l.f.Name())
+	defer l.f.Close()
 	if err := l.Writer.Flush(); err != nil {
 		return err
 	}
@@ -78,6 +80,7 @@ func (l *qlogger) Close() error {
 	if err != nil {
 		return err
 	}
+	defer f.Close()
 	buf := bufio.NewWriter(f)
 	c, err := zstd.NewWriter(buf, zstd.WithEncoderLevel(zstd.SpeedFastest))
 	if err != nil {
@@ -89,14 +92,5 @@ func (l *qlogger) Close() error {
 	if err := c.Close(); err != nil {
 		return err
 	}
-	if err := buf.Flush(); err != nil {
-		return err
-	}
-	if err := l.f.Close(); err != nil {
-		return err
-	}
-	if err := os.Remove(l.f.Name()); err != nil {
-		return err
-	}
-	return f.Close()
+	return buf.Flush()
 }
--- a/tracer.go
+++ b/tracer.go
@@ -37,10 +37,14 @@ func initQlogger(qlogDir string) logging.Tracer {
 	})
 }
 
+// The qlogger logs qlog events to a temporary file: .<name>.qlog.swp.
+// When it is closed, it compresses the temporary file and saves it as <name>.qlog.zst.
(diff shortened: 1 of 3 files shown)


tracer.go
```diff
@@ -50,50 +54,49 @@ func newQlogger(qlogDir string, role logging.Perspective, connID []byte) io.Writ
...
+	if err := c.Close(); err != nil {
+		return err
+	}
+	if err := buf.Flush(); err != nil {
+		return err
+	}
+	if err := l.f.Close(); err != nil {
+		return err
+	}
+	if err := os.Remove(l.f.Name()); err != nil {
```
marten-seemann, line 98: Isn't this error handling beautiful? 🤢

marten-seemann:
Seems to work:
Screen Shot 2021-02-11 at 18 07 28

tracer.go
```diff
@@ -50,50 +54,49 @@ func newQlogger(qlogDir string, role logging.Perspective, connID []byte) io.Writ
...
-		f:           f,
-		filename:    finalFilename,
-		WriteCloser: newBufferedWriteCloser(bufio.NewWriter(gz), gz),
+		f:        f,
+		filename: finalFilename,
+		Writer:   bufio.NewWriter(f),
 	}
 }
 
 func (l *qlogger) Close() error {
```
Stebalien, line 70: Can we add some defers for closing/deleting here? Ideally, _close_ would return after closing everything, even if something fails along the way.

Stebalien (approved):
General idea SGTM, but I want to make sure close always closes, even if we start running out of space etc.

tracer.go
marten-seemann: Makes sense, although using defer means that we won't be able to handle the error (without adding a ton of boilerplate code).

Stebalien: A helper `Compress` function to compress/copy would help a bit. Also, you may find https://pkg.go.dev/go.uber.org/multierr#Append useful.

But I don't care strongly any which way.

marten-seemann: I think it's ok that we're ignoring the error here. If closing the file fails, there's really nothing we can do about it anyway.

PR #194, opened 2021-02-19
hole punching support

vyzo:
Needs testing, but feel free to review. 

See https://github.com/libp2p/go-libp2p/issues/1048

### DIFF
--- a/listener.go
+++ b/listener.go
@@ -76,7 +76,7 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 		}
 
 		// return through active hole punching if any
-		key := holePunchKey(sess.RemoteAddr(), conn.remotePeerID)
+		key := holePunchKey{addr: sess.RemoteAddr().String(), peer: conn.remotePeerID}
 		var wasHolePunch bool
 		l.transport.holePunchingMx.Lock()
 		holePunch, ok := l.transport.holePunching[key]
--- a/listener.go
+++ b/listener.go
@@ -83,8 +83,13 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 		if ok {
 			select {
 			case holePunch.connCh <- conn:
+				// We need to delete the entry from the map here,
+				// in case we accept two connections from the same address.
+				l.transport.holePunchingMx.Lock()
+				delete(l.transport.holePunching, key)
+				l.transport.holePunchingMx.Unlock()
 				continue
-			case <-holePunch.ctx.Done():
+			default:
 			}
 		}
 
--- a/listener.go
+++ b/listener.go
@@ -77,22 +77,18 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 
 		// return through active hole punching if any
 		key := sess.RemoteAddr().String()
+		var wasHolePunch bool
 		l.transport.holePunchingMx.Lock()
 		holePunch, ok := l.transport.holePunching[key]
-		l.transport.holePunchingMx.Unlock()
-		if ok {
-			select {
-			case holePunch.connCh <- conn:
-				// We need to delete the entry from the map here,
-				// in case we accept two connections from the same address.
-				l.transport.holePunchingMx.Lock()
-				delete(l.transport.holePunching, key)
-				l.transport.holePunchingMx.Unlock()
-				continue
-			default:
-			}
+		if ok && !holePunch.fulfilled {
+			holePunch.connCh <- conn
+			wasHolePunch = true
+			holePunch.fulfilled = true
 		}
-
+		l.transport.holePunchingMx.Unlock()
+		if wasHolePunch {
+			continue
+		}
 		return conn, nil
 	}
 }
--- a/listener.go
+++ b/listener.go
@@ -12,7 +12,7 @@ import (
 
 	p2ptls "github.com/libp2p/go-libp2p-tls"
 
-	quic "github.com/lucas-clemente/quic-go"
+	"github.com/lucas-clemente/quic-go"
 	ma "github.com/multiformats/go-multiaddr"
 )
 
@@ -74,6 +74,20 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 			sess.CloseWithError(errorCodeConnectionGating, "connection gated")
 			continue
 		}
+
+		// return through active hole punching if any
+		key := sess.RemoteAddr().String()
+		l.transport.holePunchingMx.Lock()
+		holePunch, ok := l.transport.holePunching[key]
+		l.transport.holePunchingMx.Unlock()
+		if ok {
+			select {
+			case holePunch.connCh <- conn:
+				continue
+			case <-holePunch.ctx.Done():
+			}
+		}
+
 		return conn, nil
 	}
 }
@@ -92,6 +106,7 @@ func (l *listener) setupConn(sess quic.Session) (*conn, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	remoteMultiaddr, err := toQuicMultiaddr(sess.RemoteAddr())
 	if err != nil {
 		return nil, err
--- a/listener.go
+++ b/listener.go
@@ -76,7 +76,7 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 		}
 
 		// return through active hole punching if any
-		key := sess.RemoteAddr().String()
+		key := holePunchKey(sess.RemoteAddr(), conn.remotePeerID)
 		var wasHolePunch bool
 		l.transport.holePunchingMx.Lock()
 		holePunch, ok := l.transport.holePunching[key]
--- a/transport.go
+++ b/transport.go
@@ -175,7 +175,7 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 
 	if simConnect, _ := n.GetSimultaneousConnect(ctx); simConnect {
 		if bytes.Compare([]byte(t.localPeer), []byte(p)) < 0 {
-			return t.holePunch(ctx, network, addr)
+			return t.holePunch(ctx, network, addr, p)
 		}
 	}
 
@@ -225,7 +225,7 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	return conn, nil
 }
 
-func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDPAddr) (tpt.CapableConn, error) {
+func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDPAddr, p peer.ID) (tpt.CapableConn, error) {
 	pconn, err := t.connManager.Dial(network, addr)
 	if err != nil {
 		return nil, err
@@ -235,7 +235,7 @@ func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDP
 	ctx, cancel := context.WithTimeout(ctx, HolePunchTimeout)
 	defer cancel()
 
-	key := addr.String()
+	key := holePunchKey(addr, p)
 	t.holePunchingMx.Lock()
 	if _, ok := t.holePunching[key]; ok {
 		t.holePunchingMx.Unlock()
@@ -343,4 +343,8 @@ func (t *transport) Protocols() []int {
 
 func (t *transport) String() string {
 	return "QUIC"
+}
+
+func holePunchKey(addr net.Addr, id peer.ID) string {
+	return addr.String() + "__" + id.String()
 }
--- a/transport.go
+++ b/transport.go
@@ -243,6 +243,12 @@ func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDP
 	t.holePunchingMx.Unlock()
 
 	payload := make([]byte, 64)
+	var timer *time.Timer
+	defer func() {
+		if timer != nil {
+			timer.Stop()
+		}
+	}()
 	for i := 0; ; i++ {
 		if _, err := rand.Read(payload); err != nil {
 			return nil, err
@@ -255,10 +261,16 @@ func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDP
 		if maxSleep > 200 {
 			maxSleep = 200
 		}
+		d := 10*time.Millisecond + time.Duration(rand.Intn(maxSleep))*time.Millisecond
+		if timer == nil {
+			timer = time.NewTimer(d)
+		} else {
+			timer.Reset(d)
+		}
 		select {
 		case c := <-connCh:
 			return c, nil
-		case <-time.After(10*time.Millisecond + time.Duration(rand.Intn(maxSleep))*time.Millisecond):
+		case <-timer.C:
 		case <-ctx.Done():
 			t.holePunchingMx.Lock()
 			delete(t.holePunching, key)
--- a/transport.go
+++ b/transport.go
@@ -11,14 +11,13 @@ import (
 	"sync"
 	"time"
 
-	"github.com/libp2p/go-libp2p-core/connmgr"
-	n "github.com/libp2p/go-libp2p-core/network"
-
 	"github.com/minio/sha256-simd"
 	"golang.org/x/crypto/hkdf"
 
 	logging "github.com/ipfs/go-log"
+	"github.com/libp2p/go-libp2p-core/connmgr"
 	ic "github.com/libp2p/go-libp2p-core/crypto"
+	n "github.com/libp2p/go-libp2p-core/network"
 	"github.com/libp2p/go-libp2p-core/peer"
 	"github.com/libp2p/go-libp2p-core/pnet"
 	tpt "github.com/libp2p/go-libp2p-core/transport"
@@ -106,13 +105,14 @@ type transport struct {
 	gater        connmgr.ConnectionGater
 
 	holePunchingMx sync.Mutex
-	holePunching   map[string]activeHolePunch
+	holePunching   map[string]*activeHolePunch
 }
 
 var _ tpt.Transport = &transport{}
 
 type activeHolePunch struct {
-	connCh chan tpt.CapableConn
+	connCh    chan tpt.CapableConn
+	fulfilled bool
 }
 
 // NewTransport creates a new QUIC transport
@@ -153,7 +153,7 @@ func NewTransport(key ic.PrivKey, psk pnet.PSK, gater connmgr.ConnectionGater) (
 		serverConfig: config,
 		clientConfig: config.Clone(),
 		gater:        gater,
-		holePunching: make(map[string]activeHolePunch),
+		holePunching: make(map[string]*activeHolePunch),
 	}, nil
 }
 
@@ -235,26 +235,34 @@ func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDP
 	ctx, cancel := context.WithTimeout(ctx, HolePunchTimeout)
 	defer cancel()
 
-	connCh := make(chan tpt.CapableConn)
-
 	key := addr.String()
 	t.holePunchingMx.Lock()
-	t.holePunching[key] = activeHolePunch{connCh: connCh}
+	if _, ok := t.holePunching[key]; ok {
+		t.holePunchingMx.Unlock()
+		return nil, fmt.Errorf("already punching hole for %s", addr)
+	}
+	connCh := make(chan tpt.CapableConn, 1)
+	t.holePunching[key] = &activeHolePunch{connCh: connCh}
 	t.holePunchingMx.Unlock()
 
-	payload := make([]byte, 64)
 	var timer *time.Timer
 	defer func() {
 		if timer != nil {
 			timer.Stop()
 		}
 	}()
+
+	payload := make([]byte, 64)
+	var punchErr error
+loop:
 	for i := 0; ; i++ {
 		if _, err := rand.Read(payload); err != nil {
-			return nil, err
+			punchErr = err
+			break
 		}
 		if _, err := pconn.UDPConn.WriteToUDP(payload, addr); err != nil {
-			return nil, err
+			punchErr = err
+			break
 		}
 
 		maxSleep := 10 * (i + 1) * (i + 1) // in ms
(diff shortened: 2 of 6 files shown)


transport.go
```diff
@@ -257,8 +257,8 @@ func (t *transport) holePunch(network string, addr *net.UDPAddr) error {
 		return err
 	}
 
-	for i := 0; i < 5; i++ {
-		time.Sleep(time.Duration(rand.Intn(i * int(time.Millisecond))))
+	for i := 0; i < 10; i++ {
+		time.Sleep(time.Duration(rand.Intn(i * int(10*time.Millisecond))))
```
aarshkshah1992, line 261: @vyzo We already have a backoff-retry logic in the Hole Punching protocol. Do we need this ?

aarshkshah1992: Can you explain why we need this ?

vyzo: Yes, I think so; we should send multiple packets and try to coincide (within a few ms) with the inbound connection packet.

aarshkshah1992:
I don't understanding the QUIC codebase to be able to give this a meaningful review.

Requesting @marten-seemann  to please take a look.

vyzo:
I think we also need to use one of our listening sockets when doing the dial on the full dialing side as well, so that we take advantage of Cone NAT properties.

vyzo:
done.

vyzo:
oh wait, this will break assumptions about ownership of the socket; fixing.

transport.go
```diff
@@ -151,15 +155,28 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	if err != nil {
 		return nil, err
 	}
+
```
marten-seemann, line 158: ?

transport.go
```diff
@@ -151,15 +155,28 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+			}
+			return nil, err
+		}
+	}
+
 	pconn, err := t.connManager.Dial(network, addr)
 	if err != nil {
 		return nil, err
 	}
+
```
marten-seemann, line 179: ?

transport.go
```diff
@@ -202,6 +219,37 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+	if err != nil {
+		return err
+	}
+
+	_, err = conn.WriteToUDP(payload, addr)
+	if err != nil {
+		return err
+	}
+
+	for i := 0; i < 10; i++ {
```
marten-seemann, line 242: 10 seems excessive. Maybe reduce it to 3?

transport.go
```diff
@@ -202,6 +219,37 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+		time.Sleep(time.Duration(rand.Intn(i * int(10*time.Millisecond))))
+		_, err := conn.WriteToUDP(payload, addr)
```
marten-seemann, line 244: Better: generate a new random payload for each packet. Some middleboxes might inspect the packet (e.g. to detect the protocol being used) and decide to drop it if they don't like it for any reason.

transport.go
marten-seemann: Or, alternatively, should we build some cancellation logic? Something like a `map[*net.UDPAddr]context.CancelFunc activeHolePunches`, and every time `Listen` accepts a new connection it would call the cancel function if we're currently punching holes for this remote address.

transport.go
```diff
@@ -202,6 +219,37 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+
+	conn := pconn.UDPConn
+
+	payload := make([]byte, 64)
+	_, err = rand.Read(payload)
+	if err != nil {
+		return err
+	}
+
+	_, err = conn.WriteToUDP(payload, addr)
```
marten-seemann, line 237: I think you can avoid some code duplication by merging this into the loop.

transport.go
```diff
@@ -151,15 +155,28 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 	if err != nil {
 		return nil, err
 	}
 	tlsConf, keyCh := t.identity.ConfigForPeer(p)
+
+	if simConnect, _ := in.GetSimultaneousConnect(ctx); simConnect {
+		if bytes.Compare([]byte(t.localPeer), []byte(p)) < 0 {
+			err = t.holePunch(network, addr)
+			if err == nil {
+				err = errors.New("hole punching attempted; no active dial")
```
marten-seemann, line 169: Why is this an error?

transport.go
vyzo: sorry, will remove the extra newlines.

transport.go
vyzo: ditto.

transport.go
vyzo: Yeah, agreed, 10 might be a bit excessive, but I want to keep punching for a reasonable amount of time (say 100ms) until the conn is there or there is some form of timeout.

And yes indeed, having listen interrupt the hole punch if it accepts a connection from the target address is ideal.
It might be a little complex to do so, but it's the _right_ thing to do.

transport.go
vyzo: Very good point, will do.

transport.go
vyzo: Sure, will do.

transport.go
vyzo: So this is the tricky part.

Ideally we would return the connection accepted by listen, but this is problematic:
1. we need to hook into the listen and intercept that
2. the connection would be returned to the swarm twice, which is very undesirable for obvious reasons.
3. We also need to signal to the swarm that we are done trying to connect, and propagate this to the caller.

I don't think it is safe to return `nil, nil` here to the swarm, as it will probably crash if it doesn't get a connection and no error.

So I couldn't find anything better to do, so just return an error and let the caller (who actually supplied the context option after all) check for the incoming connection.

Any ideas for improving this situation are welcome!

vyzo: Granted, this should probably be a named error (if we decide to keep this logic) so that the caller knows to wait for the incoming connection and not have to guess.

transport.go
marten-seemann: I think we should also make sure to not send too many packets here. Sending a packet more often than every 10ms seems unnecessary. Maybe make it `10ms + rand.Intn(10ms)`?
That would also make sure that this function will run for (on average) 150ms instead of just 50ms.

transport.go
marten-seemann: Ok, let's do the easy thing in this PR and keep the optimization for a later time. I'd be happy to help with it btw.

transport.go
vyzo: Actually, now that I think about it, if the incoming connection succeeds the `Connect` call will not return an error to the caller.
I still think we need to return an error to the swarm (so that it doesn't expect a connect) though.

marten-seemann: Can you point me to where this `ctx` is configured and `Dial` is called? I tried grepping through the libp2p workspace, but the only occurrence for `WithSimultaneousConnect` is in go-libp2p-core, so I assume it's only in a PR at the moment?

transport.go
vyzo: thanks -- I'll check if it is reasonably easy to do right now, otherwise I will add a TODO.

transport.go
vyzo: Sure, let's not flood the network with packets.

transport.go
vyzo: It's brand new -- it's gonna come from this pr: https://github.com/libp2p/go-libp2p/pull/1057
I don't think aarsh has added yet, but he will do very soon.

transport.go
vyzo: It seems it's pretty easy to do, if we track by peer ID -- we just hook into the listener accept and cancel all the hole punching attempts for that peer.

So let's implement it, on it!

transport.go
marten-seemann: I'm ok with returning a named error here. Hooking into `Listen` and catching the connection there might be another (more complicated) solution, but I'm not sure if we're going to run into problems if both peers think that they're a client.

transport.go
vyzo: ok, I've implemented tracking with a `map[peer.ID][*net.UDPAddr]` and the listener cancels all pending hole punches when it gets a connection from the peer.

Only thing that remains is the rate (and amount of packets) that we send.
I have set a timeout of 1s and do `10ms + rand((i+1)*10*ms` for the wait.
We might want to tune this if it's too many packets.

transport.go
vyzo: Yeah, I am not sure either but I think that the swarm is not going to like it.
I'll make a named error for this, maybe it should live in core however?

vyzo: added named error.

transport.go
```diff
@@ -96,6 +103,9 @@ type transport struct {
 	serverConfig *quic.Config
 	clientConfig *quic.Config
 	gater        connmgr.ConnectionGater
+
+	holePunchingMx sync.Mutex
+	holePunching   map[peer.ID]map[*net.UDPAddr]context.CancelFunc
```
marten-seemann, line 108: Should this be a `map[*net.UDPAddr]map[peer.ID]context.CancelFunc`? 
f two peers are behind the same NAT (and therefore have the same `net.Addr`), we only need to hole punch once.

vyzo: Actually, depending on the behaviour of the NAT, we might have to hole punch twice because they will have different ports mapped.

marten-seemann: But the port is part of the `net.UDPAddr`. On the other hand, since two peers behind the same NAT will most likely use different ports, there's not much to gain from deduplicating here.

vyzo: yeah, we can't really have two peers on the same addr, port -- at least not with our code.

listener.go
```diff
@@ -92,6 +92,18 @@ func (l *listener) setupConn(sess quic.Session) (*conn, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	// cancel all active hole punches
+	l.transport.holePunchingMx.Lock()
+	holePunches, ok := l.transport.holePunching[remotePeerID]
```
marten-seemann, line 98: I'd prefer to just cancel the hole punch for this `addr`, not for all others.
The swarm will make sure that we only have a single connection to a peer, but that's a logic that should only live there. From the point of a transport implementation, it's totally fine to have multiple connections to the same peer.

Since the swarm will cancel the context passed to `Dial` as soon as it has one connection, this will also stop the hole punching for all other connection attempts.

I think this will allow you to simplify the map by removing the peer ID entry.

vyzo: Ok, can do.

vyzo: hrm, how do I get the `UDPAddr` from an `Addr`? That's all we have in the quic session.

vyzo: I guess I can use a string, assuming the two map to the same string.

vyzo: it's probably a UDPAddr underneath, but I don't really want to cast, so will use string as keys.

vyzo: done.

transport.go
```diff
@@ -202,6 +224,56 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 }
 
+func (t *transport) holePunch(ctx context.Context, p peer.ID, network string, addr *net.UDPAddr) error {
+	pconn, err := t.connManager.Dial(network, addr)
+	if err != nil {
+		return err
+	}
+	defer pconn.DecreaseCount()
+
+	ctx, cancel := context.WithTimeout(ctx, time.Second)
```
marten-seemann, line 234: Nit: put the `defer cancel()` here, not in L247.

vyzo: we can't really do that, it might have been cancelled by Listen in the meantime and we'll call it twice and panic.

transport.go
```diff
@@ -202,6 +224,56 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+
+	defer func() {
+		t.holePunchingMx.Lock()
+		defer t.holePunchingMx.Unlock()
+
+		_, ok := t.holePunching[hpkey]
+		if ok {
+			cancel()
+			delete(t.holePunching, hpkey)
+		}
```
marten-seemann, line 249: ... then this block simplifies to just the `delete`.

listener.go
```diff
@@ -79,6 +79,16 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 }
 
 func (l *listener) setupConn(sess quic.Session) (*conn, error) {
+	// cancel active hole punching if any
+	hpkey := sess.RemoteAddr().String()
+	l.transport.holePunchingMx.Lock()
+	cancel, ok := l.transport.holePunching[hpkey]
+	if ok {
+		cancel()
+		delete(l.transport.holePunching, hpkey)
```
marten-seemann, line 88: No need to `delete` here. Cleanup will be done by the `holePunch` function.

transport.go
vyzo: yes, but it will be called twice which panics!

listener.go
vyzo: we have too, so that the holePunch deferral doesn't double-cancel.

transport.go
vyzo: Note that we might want to add a limit to the number of packets (not just the timeout), so we can have two exit paths from the loop.

marten-seemann: You can call a `context.CancelFunc` as often as you want: https://golang.org/pkg/context/#CancelFunc

listener.go
marten-seemann: Yes, that should be fine.

transport.go
vyzo: oh interesting, I was under the strong impression that it would panic if you called it twice.

vyzo: ok, done.

marten-seemann (commented):
Code LGTM. Thank you @vyzo!

Once we have a test case this should be good to merge.

transport.go
aarshkshah1992: @vyzo 

```
Actually, now that I think about it, if the incoming connection succeeds the Connect call will not return an error to the caller.
```

Is there some magic in Swarm that does NOT return an error for the `Connect` call even if the dial fails but an incoming succeeds ? Please can you point me to it ?

transport.go
```diff
@@ -202,6 +224,56 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
 	return conn, nil
 }
 
+func (t *transport) holePunch(ctx context.Context, p peer.ID, network string, addr *net.UDPAddr) error {
+	pconn, err := t.connManager.Dial(network, addr)
```
aarshkshah1992, line 228: @vyzo Where is the code to stop sending packets when it sees a connection on the listener ?

vyzo: it's in the listener -- see `setupConn`.

vyzo:
what's this stuff aarsh? I am going to drop these commits when we are to merge.

aarshkshah1992:
@vyzo It's all debugging. We'll drop it all. DW

listener.go
```diff
@@ -74,6 +74,20 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
...
 			continue
 		}
+
+		// return through active hole punching if any
+		hpkey := sess.RemoteAddr().String()
+		l.transport.holePunchingMx.Lock()
+		holePunch, ok := l.transport.holePunching[hpkey]
+		l.transport.holePunchingMx.Unlock()
+		if ok {
+			select {
```
marten-seemann, line 84: There's a race condition here: If we receive 2 connections before the other go routine has deleted this entry from the `holePunching` map, we'll block here until the context is done.

marten-seemann (approved):

marten-seemann:
@Stebalien Can I ask you for a review here? I _think_ I fixed a race condition that could lead to a deadlock when we'd two connection from the same address, but I'd like to get a second set of eyes before merging this.

listener.go
Stebalien: 1. It's fixed now, right?
2. Why not delete from the hole-punching map when we first access it? That should fix all the problems here, right?

listener.go
```diff
@@ -74,6 +74,25 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 			sess.CloseWithError(errorCodeConnectionGating, "connection gated")
 			continue
 		}
+
+		// return through active hole punching if any
+		key := sess.RemoteAddr().String()
```
Stebalien, line 79: Should be `string(sess.RemoteAddr().Bytes())` (no need to convert to a _real_ string, we just need a stable key).

transport.go
```diff
@@ -202,6 +224,49 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 }
 
+func (t *transport) holePunch(ctx context.Context, network string, addr *net.UDPAddr) (tpt.CapableConn, error) {
+	pconn, err := t.connManager.Dial(network, addr)
+	if err != nil {
+		return nil, err
+	}
+	defer pconn.DecreaseCount()
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
Stebalien, line 234: Hard-coded 5s timeout?

transport.go
```diff
@@ -202,6 +224,49 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+	defer cancel()
+
+	connCh := make(chan tpt.CapableConn)
+
+	key := addr.String()
+	t.holePunchingMx.Lock()
+	t.holePunching[key] = activeHolePunch{connCh: connCh, ctx: ctx}
```
Stebalien, line 241: useless context?

transport.go
```diff
@@ -202,6 +224,49 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+	pconn, err := t.connManager.Dial(network, addr)
+	if err != nil {
+		return nil, err
+	}
+	defer pconn.DecreaseCount()
+
+	ctx, cancel := context.WithTimeout(ctx, 5*time.Second)
+	defer cancel()
+
+	connCh := make(chan tpt.CapableConn)
```
Stebalien, line 237: Size-1 buffer?

Stebalien: So we don't miss it and never block on receive.

listener.go
Stebalien: Actually, if we one-buffer the channel and atomically delete from the map, we _should_ be able to do everything under the same lock:

```go
		l.transport.holePunchingMx.Lock()
		holePunch, ok := l.transport.holePunching[key]
        if ok {
            holePunch.connCh <- conn
            delete(l.transport.holePunching, key)
        }
		l.transport.holePunchingMx.Unlock()
```

Probably not necessary, but may make things easier to reason about.

transport.go
```diff
@@ -202,6 +224,49 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+			maxSleep = 200
+		}
+		select {
+		case c := <-connCh:
+			return c, nil
+		case <-time.After(10*time.Millisecond + time.Duration(rand.Intn(maxSleep))*time.Millisecond):
+		case <-ctx.Done():
+			t.holePunchingMx.Lock()
+			delete(t.holePunching, key)
+			t.holePunchingMx.Unlock()
```
Stebalien, line 264: Check to see if _we_ succeeded in deleting it. If not, we _know_ that we're about to receive the conn, so take it. This ensures that we can't drop a connection.

transport.go
```diff
@@ -202,6 +224,49 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
+		}
+
+		maxSleep := 10 * (i + 1) * (i + 1) // in ms
+		if maxSleep > 200 {
+			maxSleep = 200
+		}
+		select {
+		case c := <-connCh:
+			return c, nil
+		case <-time.After(10*time.Millisecond + time.Duration(rand.Intn(maxSleep))*time.Millisecond):
```
Stebalien, line 260: Let's use a real timer. `time.After` leaks until the timer fires, and is significantly less efficient.

transport.go
Stebalien: Also, what if there are _two_ hole-punch attempts? We should probably error out in that case.

Stebalien (changes requested):
Mainly: make the handoff more atomic.

Stebalien:
(also happy to walk over this in-person as well)

transport.go
marten-seemann: Removed the context.
Why should two hole-punch attempts be an error?

listener.go
marten-seemann: There's no `Bytes()` method on a `net.Addr`. Also, Go has multiple ways of represent an IPv4 address, so this wouldn't be a stable key..

transport.go
marten-seemann: Introduced a `HolePunchTimeout` package-level variable.

marten-seemann:
@Stebalien
* fixed the timer
* fixed the race condition when accepting connections (this time for real, hopefully)

listener.go
Stebalien: Oh, I see. I thought this was a multiaddr.

listener.go
```diff
@@ -74,6 +74,21 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
...
+		var wasHolePunch bool
+		l.transport.holePunchingMx.Lock()
+		holePunch, ok := l.transport.holePunching[key]
+		if ok && !holePunch.fulfilled {
+			holePunch.connCh <- conn
+			wasHolePunch = true
+			l.transport.holePunching[key].fulfilled = true
```
Stebalien, line 86: We're storing these by-pointer, right? `holePunch.fulfilled = true`?

listener.go
Stebalien: Hm. Actually... we _need_ to include the peer ID here.

transport.go
Stebalien: (discussed out of band)

listener.go
marten-seemann: Yeah, I had to change that. Can't assign to a struct in a map without initialising a new struct.

listener.go
marten-seemann: Done.

transport.go
```diff
@@ -245,3 +344,7 @@ func (t *transport) Protocols() []int {
 func (t *transport) String() string {
 	return "QUIC"
 }
+
+func holePunchKey(addr net.Addr, id peer.ID) string {
```
Stebalien, line 348: nit: we can also just use a custom type `holePunchKey{addr string, id peer.ID}`. Not critical, but that would be slightly easier to reason about.

Stebalien (approved):

PR #198, opened 2021-03-19
update quic-go to v0.20.0

marten-seemann:
We'll keep using the same QUIC versions. There's no need to do a version upgrade until QUIC v1 is released.

@Stebalien v0.20.0 drops support for Go 1.14. Does this mean we need to cut a new major release? My understanding is that semver doesn't require that.


Stebalien:
> @Stebalien v0.20.0 drops support for Go 1.14. Does this mean we need to cut a new major release? My understanding is that semver doesn't require that.

That's fine. I think it's reasonable to assume users are using one of the latest two releases.

Stebalien (approved):

PR #200, opened 2021-04-02
expose some Prometheus metrics

marten-seemann:


tracer_metrics.go
```diff
@@ -0,0 +1,369 @@
...
+
+func (c *aggregatingCollector) RemoveConn(id string) {
+	c.mutex.Lock()
+	delete(c.conns, id)
+	c.mutex.Unlock()
+}
+
+var collector *aggregatingCollector
+
+func init() {
```
mxinden, line 83: If my Golang skills are not misleading me, all packages importing this package will now expose the metrics below, even if they just make use of some small part of the library, e.g. a helper function. Can the same be achieved with local instead of global variables? E.g. can one pass a metric registry at construction time and keep a reference to the metrics in some local struct?

tracer_metrics.go
```diff
@@ -0,0 +1,369 @@
...
+	const (
+		direction = "direction"
+		encLevel  = "encryption_level"
+	)
+
+	closedConns = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "quic_closed_connections",
```
mxinden, line 91:
```suggestion
			Name: "quic_connections_closed_total",
```

Small nit: There are a couple of [metric naming conventions](https://prometheus.io/docs/practices/naming/). One of them being that unit-less counters end with `_total`. Same applies to the unit-less metrics below.

tracer_metrics.go
```diff
@@ -0,0 +1,369 @@
...
+		prometheus.CounterOpts{
+			Name: "quic_new_connections",
+			Help: "new QUIC connection",
+		},
+		[]string{direction, "handshake_successful"},
+	)
+	prometheus.MustRegister(newConns)
+	bytesTransferred = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "quic_bytes_transferred",
```
mxinden, line 107: Another naming convention is to move the unit to the end of the metric name:

```suggestion
			Name: "quic_transferred_bytes",
```

mxinden (commented):
Great to see instrumentation for QUIC to come in the future!

tracer_metrics.go
```diff
@@ -0,0 +1,369 @@
...
+func (c *aggregatingCollector) Describe(descs chan<- *prometheus.Desc) {
+	descs <- c.rtts.Desc()
+	descs <- c.connDurations.Desc()
+}
+
+func (c *aggregatingCollector) Collect(metrics chan<- prometheus.Metric) {
+	now := time.Now()
+	c.mutex.Lock()
+	for _, conn := range c.conns {
+		if rtt, valid := conn.getSmoothedRTT(); valid {
```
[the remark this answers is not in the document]
marten-seemann: There are 2 problems with recording the RTT every roundtrip:
1. If a connection is not sending any data, it will not generate any RTT samples.
2. If we have one connection with 50ms RTT and one connection with 500ms RTT (assuming both are sending data and both are using the same congestion window), the shorter-RTT connection will generate 10x as many RTT measurements.

What I'm trying to measure is the RTT distribution of all active connections, so we have to make sure that every connection generates the same number of measurements per scraping interval.

tracer_metrics.go
marten-seemann: What's the best practice here? If `package` wants to expose Prometheus metrics, does the user of that package call a `package.InitializeMetrics()` function? We could certainly do that, if that's the preferred way.

tracer_metrics.go
marten-seemann: Thanks, I renamed most of the metrics.

tracer_metrics.go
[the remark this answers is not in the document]
mxinden: > If a connection is not sending any data, it will not generate any RTT samples.

This applies to both the status quo and my suggestion, correct?

> If we have one connection with 50ms RTT and one connection with 500ms RTT (assuming both are sending data and both are using the same congestion window), the shorter-RTT connection will generate 10x as many RTT measurements.

Good point. Thanks for clarifying!

> What I'm trying to measure is the RTT distribution of all active connections, so we have to make sure that every connection generates the same number of measurements per scraping interval.

Would a collection across connections every X seconds do the trick? Thus data collection would not depend on external factors (number of Prometheus instances and scrape interval) but at the same time uphold the same amount of measurements per connection.

tracer_metrics.go
mxinden: > What's the best practice here?

If I recall correctly the best practice in Golang is to use global metric variables and initialization via `init` like you did here. I am not in favor of this instrumentation style, due to the various issues that global state brings along. Instead I would pass a registry as an argument at object creation, e.g. `NewQuicTransport`. The returned `QuicTransport` object would then own the metrics registered with the provided Registry.

I am happy to expand on my architecture, though in the end I think you should go with whatever works best for you.

marten-seemann: > Instead I would pass a registry as an argument at object creation, e.g. NewQuicTransport. 

What would happen if you create multiple transports then? We can only register each metric once.

If the `init` way is actually the common way to do this in Go, let's get this PR merged. We can always improve it later on.

tracer_metrics.go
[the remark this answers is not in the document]
marten-seemann: > This applies to both the status quo and my suggestion, correct?

We're saving the most recent RTT sample for every connection, so this problem is solved in the PR. We might use an RTT measurements that a few seconds old, but considering that we're sending a keep-alive PING every 15s, we're generating fresh RTT measurements quite frequently.

> Would a collection across connections every X seconds do the trick? Thus data collection would not depend on external factors (number of Prometheus instances and scrape interval) but at the same time uphold the same amount of measurements per connection.

What's the advantage of that? Looping over all connections is moderately expensive, so it would be good to not do it more frequently than necessary.

tracer_metrics.go
mxinden: > What would happen if you create multiple transports then? We can only register each metric once.

What I would do is have a `RegisterMetrics` function that takes a registry and returns a structure with all the metrics. I would then pass a clone of the latter to each transport. 

> If the init way is actually the common way to do this in Go, let's get this PR merged. We can always improve it later on.

That sounds good to me.

tracer_metrics.go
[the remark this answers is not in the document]
mxinden: > What's the advantage of that?

Looping over the connections on a fixed interval would make the data returned to Prometheus independent of its scrape interval and the amount of Prometheus instances. E.g. if you have 3 Prometheus instances for high availability, you would be looping over the connections 3 times every scrape interval (e.g. minute).

I think it is fine either way.

tracer_metrics.go
```diff
@@ -0,0 +1,369 @@
...
+		prometheus.CounterOpts{
+			Name: "quic_packets_dropped_total",
+			Help: "Dropped packets",
+		},
+		[]string{"packet_type", "reason"},
+	)
+	prometheus.MustRegister(droppedPackets)
+	connErrors = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "quic_connection_errors",
```
mxinden, line 147:
```suggestion
			Name: "quic_connection_errors_total",
```

Nit pick: This is a unit-less counter and should thus be suffixed with `_total`.

mxinden (approved):
Looks good to me from the Prometheus side. Not sure my Golang skills are good enough for a general approval. Your call.

tracer_metrics.go
marten-seemann: Fixed.

tracer_metrics.go
marten-seemann: > What I would do is have a RegisterMetrics function that takes a registry and returns a structure with all the metrics. I would then pass a clone of the latter to each transport.

That sounds like a much cleaner setup. I'm open to making this change in the future.

PR #210, opened 2021-07-05
don't compare peer IDs when hole punching

marten-seemann:
@mxinden pointed out that comparing peer IDs when hole punching is not necessary. We'll have logic in our DCUtR protocol anyway to assign the role of client and server (we'll need the same logic to coordinate the TCP sim open).

### DIFF
--- a/conn_test.go
+++ b/conn_test.go
@@ -377,7 +377,7 @@ var _ = Describe("Connection", func() {
 		connChan := make(chan tpt.CapableConn)
 		go func() {
 			defer GinkgoRecover()
-			conn, err := t2.Dial(n.WithSimultaneousConnect(context.Background(), ""), ln1.Multiaddr(), serverID)
+			conn, err := t2.Dial(context.Background(), ln1.Multiaddr(), serverID)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


vyzo:
Sounds like a good idea, but we'll have to change the code in the hole punching pr in go-libp2p so that only one of the peers uses the option, predictably.

marten-seemann:
I added a comment to that PR, so we don't forget. I suggest moving forward with this PR, and not blocking on that change.

vyzo:
SGTM

mxinden (approved):
> @mxinden pointed out that comparing peer IDs when hole punching is not necessary. We'll have logic in our DCUtR protocol anyway to assign the role of client and server (we'll need the same logic to coordinate the TCP sim open).

Approving the idea of assigning roles through the DCUtR protocol instead of by comparing peer IDs.

Stebalien (approved):
Someday, maybe, we can expose a separate `HolePunch` (or `ActiveListen`) function instead....

marten-seemann:
> Someday, maybe, we can expose a separate HolePunch (or ActiveListen) function instead....

I wouldn't be opposed to that.

PR #211, opened 2021-07-07
implement a Transport.Close that waits for the reuse's GC to finish

### LINKED ISSUE #176: implement transport.Close()
Once https://github.com/libp2p/go-libp2p/issues/999 is merged.

marten-seemann:
This finally makes it possible to run tests with race detector, and therefore switch to the Unified CI Setup.

Not sure if the implementation of `Transport.Close` here is sufficient to fix #176 though.

### DIFF
--- a/reuse.go
+++ b/reuse.go
@@ -5,8 +5,6 @@ import (
 	"sync"
 	"time"
 
-	"github.com/libp2p/go-libp2p-core/connmgr"
-
 	"github.com/libp2p/go-netroute"
 )
 
@@ -24,7 +22,7 @@ type reuseConn struct {
 	unusedSince time.Time
 }
 
-func newReuseConn(conn *net.UDPConn, gater connmgr.ConnectionGater) *reuseConn {
+func newReuseConn(conn *net.UDPConn) *reuseConn {
 	return &reuseConn{UDPConn: conn}
 }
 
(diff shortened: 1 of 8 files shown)


reuse.go
```diff
@@ -204,3 +210,14 @@ func (r *reuse) Listen(network string, laddr *net.UDPAddr) (*reuseConn, error) {
 	r.unicast[localAddr.IP.String()][localAddr.Port] = rconn
 	return rconn, err
 }
+
+func (r *reuse) Close() error {
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
+	close(r.closeChan)
+	if r.garbageCollectorRunning {
+		<-r.garbageCollectorStopChan
```
Stebalien, line 219: This can deadlock if we're waiting for the garbage collection routine to take the lock.

Stebalien (changes requested):

marten-seemann:
I removed the rather complicated logic of starting and stopping the reuse gc function. Now that we have a clean shutdown procedure, this won't be necessary any more.

Stebalien (changes requested):
> I removed the rather complicated logic of starting and stopping the reuse gc function. Now that we have a clean shutdown procedure, this won't be necessary any more.

We're still stopping the garbage collector when we have no more connections. We need to remove that if we're going to remove the auto start/stop feature.

marten-seemann:
> We're still stopping the garbage collector when we have no more connections. We need to remove that if we're going to remove the auto start/stop feature.

Oops, must have missed that. Fixed.

Stebalien (approved):

Stebalien:
LGTM but there's a test failure we should probably investigate.

marten-seemann:
> LGTM but there's a test failure we should probably investigate.

Looks unrelated, I opened #213.

PR #239, opened 2021-10-27
chore: update deps

aschmahmann:
@marten-seemann I bumped a bunch of the dependencies here. Most of these are dependencies that are already present in places like go-libp2p, but there's not necessarily a pressing need for them.

Figured this might be an illustrative way of looking at our dependency graph and how it falls behind.


marten-seemann (approved):
-266 LOC in `go.sum`? That's amazing, looks like one of the dependencies cleaned up its dependency tree.

marten-seemann:
Reverting, since it broke the build on master.

marten-seemann:
Weird, trying to recreate this PR leads to massively inflating `go.sum` and `go.mod`.

PR #249, opened 2021-12-31
use the Resource Manager

marten-seemann:
This is without passing `ReserveMemory` functions to quic-go. We're just using it to open / accept connections and streams.

When reviewing, please pay attention to the hole punching logic.


conn.go
```diff
@@ -38,14 +40,28 @@ func (c *conn) IsClosed() bool {
...
 func (c *conn) AcceptStream() (mux.MuxedStream, error) {
 	qstr, err := c.sess.AcceptStream(context.Background())
-	return &stream{Stream: qstr}, err
+	if err != nil {
+		return nil, err
+	}
+	if _, err := c.rcmgr.OpenStream(c.remotePeerID, network.DirInbound); err != nil {
+		return nil, err
+	}
+	return &stream{Stream: qstr}, nil
```
vyzo, line 64: shouldnt it keep the stream scope?

vyzo: we leak it otherwise

listener.go
```diff
@@ -70,7 +75,13 @@ func (l *listener) Accept() (tpt.CapableConn, error) {
 			sess.CloseWithError(0, err.Error())
 			continue
 		}
-		if l.transport.gater != nil && !(l.transport.gater.InterceptAccept(conn) && l.transport.gater.InterceptSecured(n.DirInbound, conn.remotePeerID, conn)) {
+		connScope, err := l.rcmgr.OpenConnection(network.DirInbound, false)
```
vyzo, line 78: where is this passed?

transport.go
```diff
@@ -183,17 +191,25 @@ func (t *transport) Dial(ctx context.Context, raddr ma.Multiaddr, p peer.ID) (tp
...
 	tlsConf, keyCh := t.identity.ConfigForPeer(p)
-
-	if ok, isClient, _ := n.GetSimultaneousConnect(ctx); ok && !isClient {
-		return t.holePunch(ctx, network, addr, p)
+	if ok, isClient, _ := network.GetSimultaneousConnect(ctx); ok && !isClient {
+		return t.holePunch(ctx, netw, addr, p)
 	}
 
-	pconn, err := t.connManager.Dial(network, addr)
+	scope, err := t.rcmgr.OpenConnection(network.DirOutbound, false)
```
vyzo, line 198: and here too, where is the scope passed?

vyzo (commented):
we'll need to hook the memory rsvp as well, but ok for now.

listener.go
marten-seemann: Fixed by using the new `transport.CapableConn` interface.

transport.go
marten-seemann: Fixed by using the new `transport.CapableConn` interface.

conn.go
marten-seemann: The `MuxedStream` doesn't have any scope.
This is probably the wrong layer to do it, this should be done in the swarm.

marten-seemann: This is now done in the swarm, see https://github.com/libp2p/go-libp2p-swarm/pull/308.

vyzo (commented):
we need to Done the scope on closure

listener.go
```diff
@@ -102,20 +105,23 @@ func (l *listener) setupConn(sess quic.Session) (*conn, error) {
...
 	remotePeerID, err := peer.IDFromPublicKey(remotePubKey)
 	if err != nil {
 		return nil, err
 	}
-
 	remoteMultiaddr, err := toQuicMultiaddr(sess.RemoteAddr())
 	if err != nil {
 		return nil, err
 	}
+	connScope, err := l.rcmgr.OpenConnection(network.DirInbound, false)
```
marten-seemann, line 116: Should we loop here, or return the error to the caller?

vyzo: let's bail.

vyzo (approved):

PR #250, opened 2022-01-03
migrate to standard Go tests, stop using Ginkgo

marten-seemann:
Paying off technical debt. This will allow us to move faster in the future, especially when it comes to repo consolidation.


vyzo (approved):
woah, whats this weird testing framework!

PR #252, opened 2022-01-07
don't start a Go routine for every connection dialed

marten-seemann:
In the swarm, we're calling Close for every connection before it is garbage collected. We therefore don't need to start a Go routine here just to see when a connection is closed.
We now also increment the reuse counter every time a connection is dialed. This simplifies closing the connection.


vyzo (approved):
nice, no need to spawn that goroutine.

PR #272, opened 2022-05-04
feat: add `DisableReuseport` option

### LINKED ISSUE #1428: quic: make the reuseport feature optional
Hi from @berty,
We recently discovered that if we enable the `reuseport` feature on the `libp2p-tcp-transport` while using a cellular connection, it could lead to network errors with some mobile operators. This is the case with the French operator Free Mobile which is massively used in France and we suspect that the problem occurs with several other operators around the world.

It seems that the QUIC transport has exactly the same issue but does not have an option to disable the `reuseport`. At the moment, this prevents us from using the QUIC transport on cellular connection even though it is particularly suitable for mobile use.

It would be nice to make the feature optional like in the `libp2p-tcp-transport`. I checked if it was possible to simply add an option to be able to toggle it off, but the feature seems too tightly bonded to the rest of the code.

gfanton:
Fixes https://github.com/libp2p/go-libp2p/issues/1428

Hey, I just started to add the `DisableReuseport` options.  It seems to be working as expected with some cellular operators that couldn't maintain the connection (I tried with this script https://github.com/gfanton/libp2p-reuseport-test)

Just some few things:

- It seems that there is no option/configuration mechanism, so I added one. Let me know if you are ok with this or if you want to use another method to pass the argument.
- I wrapped all the `conn_tests` with a simple table test case to test everything conn related with the reuseport_on/off
- I have to use `net.ListenUDP` instead of `net.DialUDP` for the Dial method to work, otherwise the call to the `WriteTo` method fails.

Also there are two tests that currently fail with the reuseport disabled:

**TestHolePunching**
```
=== RUN   TestHolePunching/reuseport_off
    conn_test.go:59: using a Secp256k1 key: 16Uiu2HAmPxEseg6Hti88hpHTyUB3tQk9bw7paj9yS3rPkBgPnWgM
    conn_test.go:59: using a Ed25519 key: 12D3KooWAaU8sc9j783H9hH3eG3yyJDnvz9Le1LdhCQFbYe5gFyu
    conn_test.go:581: 
        	Error Trace:	conn_test.go:581
        	           				asm_arm64.s:1133
        	Error:      	An error is expected but got nil.
        	Test:       	TestHolePunching/reuseport_off
        	Messages:   	didn't expect to accept any connections
    conn_test.go:602: 
        	Error Trace:	conn_test.go:602
        	Error:      	Condition never satisfied
        	Test:       	TestHolePunching/reuseport_off
```

**TestStatelessReset**
```
=== RUN   TestStatelessReset/reuseport_off
    conn_test.go:59: using a ECDSA key: QmRoF9AeMPnxYMrErq91y2xCvjiULp9n9cJuVwDHu9Kdao
    conn_test.go:59: using a RSA key: QmNvwzR7VDFy1wFdHE3PgV8mCztJbFWz1d5NXVL3wXYVgX
    conn_test.go:534: 
        	Error Trace:	conn_test.go:66
        	           				conn_test.go:534
        	Error:      	Received unexpected error:
        	           	listen udp4 127.0.0.1:55213: bind: address already in use
        	Test:       	TestStatelessReset/reuseport_off
```

Do these tests make sense when used with resuseport disabled?

### DIFF
--- a/conn.go
+++ b/conn.go
@@ -2,6 +2,7 @@ package libp2pquic
 
 import (
 	"context"
+	"net"
 
 	ic "github.com/libp2p/go-libp2p-core/crypto"
 	"github.com/libp2p/go-libp2p-core/network"
@@ -12,9 +13,17 @@ import (
 	ma "github.com/multiformats/go-multiaddr"
 )
 
+type pConn interface {
+	net.PacketConn
+
+	// count conn reference
+	DecreaseCount()
+	IncreaseCount()
+}
+
 type conn struct {
 	quicConn  quic.Connection
-	pconn     *reuseConn
+	pconn     pConn
 	transport *transport
 	scope     network.ConnManagementScope
 
--- a/conn_test.go
+++ b/conn_test.go
@@ -30,6 +30,16 @@ import (
 
 //go:generate sh -c "mockgen -package libp2pquic -destination mock_connection_gater_test.go github.com/libp2p/go-libp2p-core/connmgr ConnectionGater && goimports -w mock_connection_gater_test.go"
 
+type connTestCase struct {
+	Name    string
+	Options []Option
+}
+
+var connTestCases = []connTestCase{
+	{"reuseport_on", []Option{}},
+	{"reuseport_off", []Option{DisableReuseport()}},
+}
+
 func createPeer(t *testing.T) (peer.ID, ic.PrivKey) {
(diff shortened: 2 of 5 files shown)


gfanton:
didn't see https://github.com/libp2p/go-libp2p/issues/1187, Im going to move this on `go-lib2p` repo

